PR #1, opened 2015-01-27
Add data for Fine Offset sensor

vestom:
Manually created from rtl_433
Needs -m 1 when reading


merbanan:
Can you submit another signal? One that can be used with -m 0, I have updated the instructions.

PR #2, opened 2015-01-29
A couple of Oregon Scientific sensors

helgew:
I named the files according to each sensor's model number... not sure if that would work with the testing framework you have in mind. Should we also submit what the data output should look like?


merbanan:
In the future the test framework will figure out how the output should look like.

merbanan:
Thanks for the data.

PR #3, opened 2015-01-29
Add Owl CMR180 test file

onlinux:
Hello,
I don't know if it is the good way to send test files, but i tried to understand the instructions the best i could.


merbanan:
You should read up on git rebase --interactive. You can then fix your commits so they look nice then. But I'll merge this as it doesn't matter.

onlinux:
Thank you very much for this hint.
Actually, i am quite new with git and i have to admit that is not very user
friendly.

2015-01-29 20:09 GMT+01:00 Benjamin Larsson <notifications@github.com>:

PR #4, opened 2015-01-30
Added data files for the two types of transmissions from VN1TXC sensors

helgew:
These were saved while a second dongle reported the data parsed from the VN1TXC sensor. However, they are currently not parsed correctly when reading them in again with the '-r' command line flag.


merbanan:
Thanks

PR #11, opened 2015-06-30
Add LaCrosse TX6U, TX7U, and TX22U sample recordings

rct:


rct:
I've added sample files for four different device types.   However,  I didn't (neglected to) create a separate branch for each one.

* LaCrosse TX 433 Mhz
* LaCrosse TX22 915 Mhz
* X10 RF power control from PalmPad Remote
* DSC alarm system door/window security contact.

PR #13, opened 2015-07-12
Add test script and delete empty file

vestom:


merbanan:
We need to add some kind of testing of the output later. Anyway great job.

vestom:
It will probably be a huge effort to verify the output - especially as it is so diverse between drivers.
By piping output from the script to a file and comparing before and after it is easy to catch regressions. Example (on linux):
$ rtl_433_test.sh &>after.txt
$ meld before.txt after.txt

merbanan:
What I want is something that can track changes. So that one can know if one change breaks some feature.

PR #20, opened 2015-08-16
test files for tfa twin plus 30.3049 thermo/hygrometer

iotzo:
test files for tfa twin plus 30.3049 thermo/hygrometer


merbanan:
Thanks

PR #21, opened 2015-08-19
Test data from waveman switch protocol recorded from Intertechno ITK-…

ortylp:
…200 Model KCT-510

All possible switch positions recorded.


merbanan:
Thanks

PR #25, opened 2015-08-23
acurite model 00606TX

jayallenpdx:
test data and analysis for Acu-Rite model 00606TX temperature only sensor.


merbanan:
thanks

PR #27, opened 2015-08-26
Update sample files for Alectov1 protocol

jcarduino:
HEre are example files for alectov1 protocol. also included pdf with protocol details and readme.md with description of files


merbanan:
Thanks

PR #34, opened 2015-09-01
Added samples and details for Auriol H13726A,LD3091 and Silvercrest Z30914

xes:
Added samples and details for Auriol H13726A,LD3091 and Silvercrest Z30914


merbanan:
Thanks

PR #36, opened 2015-09-05
Add samples for the digitech weather station

viraptor:
Add samples and an emergency mechanism to tolerate bad quality samples.


merbanan:
thanks

PR #40, opened 2015-10-17
Add test files for new device Efergy e2 classic

robho:


merbanan:
Ops missed this. Thanks.

PR #43, opened 2015-11-21
Acurite 986 Refrigerator and Freezer Thermometer

rct:
Two test data files, covering positive and negative temperatures, as well as low battery.


merbanan:
Thanks.

PR #44, opened 2015-11-21
Add "prologue" data+pictures

enavarro222:
Small contribution. Add data and pictures for a cheap "prologue" compatible sensor


merbanan:
Thank you.

PR #51, opened 2015-12-06
Script to run automatic tests

enavarro222:
Modest script to run automatic test on repository sample data. It looks for all `.json` files under the `tests` dir and compare it to rtl_433 output ran over same file with `.data` ext.

It depends on [deepdiff](http://deepdiff.readthedocs.org/), to install it:

    $ pip install deeddiff

To run the tests:

    $ python bin/run_test.py -I time

The option `-I time` is to ignore "time" field. Something  better may be done to manage time check...

One can also specify `rtl_433` path:

     $ python bin/run_test.py -c "../build/src/rtl_433" -I time

This may be largely improved. However I prefer to share quickly.

The PR also contains some expected json samples.

Also, I see that README.md refer to a "make test" however I didn't find any makefile...

### DIFF
--- a/bin/run_test.py
+++ b/bin/run_test.py
@@ -33,7 +33,7 @@ def remove_fields(data, fields):
 
 def main():
     parser = argparse.ArgumentParser(description='Test rtl_433')
-    parser.add_argument('-c', '--rtl433-cmd',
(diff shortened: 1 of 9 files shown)


merbanan:
Thanks

PR #55, opened 2015-12-10
Add failing test capture

dwmw2:
rtl_433 is working on the others now. Not this one though.


dwmw2:
Not quite what I was looking for in https://github.com/merbanan/rtl_433/issues/227 but interesting nonetheless...

PR #57, opened 2015-12-10
Add another failing test capture for oil_watchman

dwmw2:
When I left rtl_433 do the FM decode, my own pulse decoder can process the output and find the data. rtl_433 can't though.


dwmw2:
This one can be decoded if as well as setting `-l2000` I also reduce FSK_DEFAULT_FM_DELTA to 4000.

Perhaps that should be proportional to `s->fm_f1_est`, rather than an absolute value? In fact, when we build up `s->fm_f1_est` at the start of the pulse, how about we also calculate the stability? If we later see a value which is 4 times further away than the delta between our observed minimum and maximum, that's a fairly clear hint that the pulse just ended :)

merbanan:
I have the pleasure of maintaining code that someone else wrote so most of the things you say sounds nice. But what you are describing is some kind of agc. I've had some ideas on how to do that for the ook demod but just setting a lower default made it good enough. For amplitude modulated signals the actual rtl-sdr hardware has automatic gain. If you find some heuristics that works better and is simple I am all for it.

dwmw2:
Reducing the default works for me in the short term, certainly. It looks like we have a good set of test samples in rtl_433_tests so we can test whether it would break things for anyone else, which is good. But I haven't yet worked out how to run the tests automatically.

I'm a little dubious about why we *need* AGC for the FM signal. Surely a given value the FM demodulation should indicate a specific frequency. I don't understand why it's acceptable for that level to *vary* according to the amplitude of the incoming signal. It's *not* amplitude modulation; it's supposed to be frequency modulation! So I suspect that fixing the FM demod as discussed in https://github.com/merbanan/rtl_433/issues/223 would actually make this problem go away? And that implementing AGC on the FM-demodulated signal would just be a poor workaround?

I do think that ideally, we want each pulse stream to have a kind of 'credibility' attached to it, according to how strong the signal was and how clean the edges were. And a low-credibility signal should *only* be offered to device decoders with a strong CRC which are unlikely to accept something that isn't really theirs.

dwmw2:
Of the signals I've collected in rtl_433_tests so far, I can decode them all by halving FSK_DEFAULT_FM_DELTA and playing with the (AM) detection level. However, there is no single detection level which works for all the samples. And why should there be? This is FM, used precisely so that we don't *need* to have a strong signal. I think perhaps we should decouple the FM pulse detection from the AM pulse detection a little more. One of the other samples *starts* being picked up with `-l 2000` but then stops half-way and is detected as a gap, even though the FM data are completely clear still.

merbanan:
This limit is just a threshold value. For the OOK demod it is connected to the amplitude of the signal. For the FSK it is not. But both demodulates demodulate to an envelope that is processed the same way. To start with maybe just add a separate limit value for the am and fsk demod.

dwmw2:
Just to be clear: in the current implementation, the FSK demod *is* connected to the amplitude of the signal. We don't even start to do FSK processing until we've actually detected the first OOK pulse and gone from `PD_STATE_IDLE` to `PD_STATE_FIRST_PULSE` *(and also `PD_STATE_FSK_F1`)*. And even then, if the *AM* signal value goes below the detection limit, we stop processing the FSK frequency data and go back to `PD_STATE_GAP`.

But yes, decoupling those and being able to process the FM-demodulated data on its own does seem to make sense.

rct:
FYI, only 5 people watch the rtl_433_tests repo, you aren't going to get much feedback.

@vestom wrote the FM code, he's not watching this repo, at the moment.

You might consider moving the discussion to the rtl_433 google group, or at least an issue on rtl_433 instead of rtl_433_tests.

vestom:
Hi,
As Robert mentioned, I am the one to blame for the FSK demodulation :-) As mentioned in the rtl_433 wiki, discussion should probably take place in https://groups.google.com/forum/#!forum/rtl_433
At least it is the only place I regularly check...

The idea behind the FSK demodulation is that it uses the OOK detection for carrier detection. To avoid too much noise and false trigging, a stable carrier in the envelope signal should be present, corresponding to `PD_STATE_FIRST_PULSE`. If the carrier disappears, why should be continue FSK demodulation?

I agree the current OOK level detection is very crude, and I have actually been toying with the idea of making the OOK level detection adaptive by using noise and level estimators. That would make OOK threshold detection better for both noisy signals and weaker signals.

The FSK uses two frequency estimators (`s->fm_f1_est` and `s->fm_delta_est`), as the FSK offsets are not known in advance and may change with centre frequency etc. Delta is the difference between frequency f1 and f2. `FSK_DEFAULT_FM_DELTA` is the default delta estimate and thereby setting the minimum frequency difference for FSK decoding. (See also the discussion here: https://groups.google.com/forum/#!topic/rtl_433/n7Eav-a5LKU)

I do not agree it would be a good idea to make `FSK_DEFAULT_FM_DELTA` proportional to `s->fm_f1_est`, as `s->fm_f1_est` is dependant on the offset from the tuner frequency whereas `FSK_DEFAULT_FM_DELTA` is dependant on the sampling frequency. 

The FSK decoder is quite new and immature, but the default threshold seemed to work with the two sensors I could test. The pulse detector state machine is also becoming rather unwieldy and could probably be refactored. I would be happy to discuss it further and hopefully make it even better.

I would also like to have a look at your troublesome signals. But please open a new topic at https://groups.google.com/forum/#!topic/rtl_433 and let us continue the discussion there.

/Tommy

PR #59, opened 2015-12-12
Improve run_test script

enavarro222:
Here is some small improvement of run_test script (still a lot to do):
* can handle "multi-lines" i.e. samples that output more than one data packet. It's not activated from makefile as for many devices, there is multiple lines output as bitbuffer rows are not well captured.
* print a short summary of the test at the end

I also add a makefile to run it easily.


merbanan:
Thanks.

PR #60, opened 2015-12-12
Othe transmission from Watchman sonic unit

fraschizzato:
Those are the transmissions from 2 different watchman sonic unit. Actually that transmission where not decoded from rtl_433, even specifing -R 42.
Note: probably there are transmission from other devices in the captured data, there's an efergy e2 classic in the building (not mine)...


merbanan:
Well when decoding works we can move the signals to the correct directory.

PR #61, opened 2015-12-12
Oregon Scientific Protocol 1.0

ygator:
The transmission is repeated.

0198821b

Nibble 0 = CC??
CC = Channel - 1
?? = ??

Nibble 4 = B?S?
B = Low Battery
? = 
S = Temperature Sign 0=+ 1=-
? =

Nibble 1 = Power On ID

Nibble 5 Nibble 2 . Nibble 3 = BCD Temperature in Celsius

Nibble 6 Nibble 7 = Byte 3 = Checksum
Checksum equals Sum of Bytes 0 1 2

So this transmission is
Channel 1
Battery OK
Power On ID = 1
Temperature +29.8C


merbanan:
Can you add a few notes about this in a text file.

PR #75, opened 2016-01-27
Generic door sensor tests

rbadsi:
Hi,

My branch includes test files for a generic door sensor.

Thanks

Radim


merbanan:
Looks good, thanks.

rct:
If you submit a decoder, perhaps name it `hs1527_*`  since you know the chip being used.  That's probably the best identifier for now.

I think after we get enough "no-name" products all named `generic_*` it will cause unnecessary confusion.

FWIW, looks like there is Arduino code for the x1527:
* http://forum.arduino.cc/index.php?topic=99714.0
* http://tinkerman.eldiariblau.net/decoding-433mhz-rf-data-from-wireless-switches-the-data/

I suspect there will be a number of devices that will have an x1527 in it though with different timings.

Unfortunately at first glance it doesn't look like there is checksum or crc, so it's ripe for false positives. 

(I sent this before during the github outage, hopefully it won't get posted twice.)

PR #76, opened 2016-01-31
Created fineoffset_wh1080.c

ovrheat:
This module is based on Stanisław Pitucha ('viraptor' https://github.com/viraptor ) earl code for the Digitech XC0348 Weather Station, which seems to be a rebranded Fine Offset WH1080 Weather Station. The latter is the 'real' product, rebranded in many ways (see comments in fineoffset_wh1080.c ), so I guess that this module should substitute the earl Digitech XC0348 (still incomplete) Stanisław's code. 
See  the product page here: http://www.foshk.com/weather_professional/wh1080.htm . 
The module seems working fine and *should* (I hope) work with most of the brands/models of the WH1080. It actually performs decoding on all of the sensors data (temperature, humidity, wind speed and direction, rain) and also time signal (DCF77) .
In my repository https://github.com/ovrheat/rtl_433.git you will find an rtl_433 snapshot with this module working, I just have removed Digitech XC0348 because of conflicts.
I hope you like it! :)

### DIFF
--- a/tests/Fineoffset_WH1080/fineoffset_wh1080.c
+++ b/tests/Fineoffset_WH1080/fineoffset_wh1080.c
@@ -1,0 +1,311 @@
+/*
+ * *** Fine Offset WH1080 Weather Station ***
+ * (aka Watson W-8681)
+ * (aka Digitech XC0348 Weather Station)
+ * (aka PCE-FWS 20) 
+ * (aka Elecsa AstroTouch 6975)
+ * (aka Froggit WH1080)
+ * (aka .....)
+ *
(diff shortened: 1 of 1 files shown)


merbanan:
Are there any test data available in the rtl_433_tests repository for this decoder ? If not you need to supply that before I can merge this decoder. Can you post a link where one could buy this hardware as DCF77 support might be of interest.

merbanan:
Ok I now see this is in the wrong repository, can you resend it to the rtl_433 repo ?

merbanan:
Or more correct, can you merge it to the other decoder and complete the tests repository with data files from your sensor. As this does not belong in this repository I well close the ticket.

PR #77, opened 2016-02-02
Updated Fine Offset WH1080 test directory

ovrheat:
This is the test suite for the Fine Offset WH1080 weather station. Inside its folder (fineoffset_wh1080) there are gdata files for weather and time (DCF77) decoding. 
I hope that everything is fine now, I'm not much familiar with Github.


merbanan:
look good

PR #88, opened 2016-04-09
Test data from my Apollo watchman

tomq42:
This doesn't appear to be recognised by rtl_433.


tomq42:
The captures are the result of putting a magnet next to the black dot.

PR #112, opened 2016-08-31
Acurite "TH" 609A1TX Temperature and Humidity sensor

rct:
Adding missing samples for device 11 - "Acurite Temperature and Humidity Sensor", aka acurite_th. 

Not sure who originally implemented this, but there were no samples in rtl_433_tests.

Samples include both positive and negative temperatures.

TODO - low battery message if it sends one.


merbanan:
Thanks

PR #113, opened 2016-08-31
"Corradi" Remote Controll

socrat3:
Data file for Corradi Remote controll with 15 channels.

I have recorded the 5 channel for button up, stop and down.


merbanan:
Thanks

socrat3:
i dont understand this code. 

I own a Pergotenta (rolling pergola) by an italian company called Corradi. Their remote control is microchip based on 433MHz. The remote's commands are decoded by the RFCCom when KeeLoq is activated I understand there is both a clear part and an encrypted part. In RFXmngr there is no sigh of the encrypted part though so I am not sure they actually use it.

rtl_433 support keeloq or rolling code system?

merbanan:
There is no support for keeloq or any rolling code yet. And for decoding of any system with encrypted parts you need the decryption key and algorithm to decrypt it.rtl_433 has none of that yet. Patches welcome though. We can help with decoding the non encrypted parts though.

socrat3:
do you know https://github.com/twmw/keelog-python?

merbanan:
No, if you produce a signal sample and the corresponding encryption key and necessary documentation I will add an api where you can add encryption keys so you can use them to decode the encrypted data for such signals.

PR #121, opened 2016-10-02
Acurite 6045 lightning detector samples

rct:
Added some initial samples of Acurite 06045 lightning detector samples.

Will need to add a more diverse set later.


merbanan:
Sorry I missed this pull. Thanks for the samples.

PR #123, opened 2016-10-16
Honeywell (Ademco) Window/Door Security Sensor

adam1010:
Tests to accompany the new Honeywell driver in the rtl_433 repository


merbanan:
If there are any other written specs you can send it here also.

PR #132, opened 2017-01-07
added detected signals

klattimer:
A collection of my RF 433Mhz devices


klattimer:
I have more of these to come, Kerui PIR motion sensors, especially as my first pass at support doesn't work with ALL of my sensors, one is generating an erroneous signal and I need to catch why. 

I also have some cheapo LED switches, a doorbell and will be acquiring some other bits and pieces shortly.

PR #137, opened 2017-01-20
beckett rocket tek377a

mattrivett:
I'm not sure if I'm doing the pull requests right but here goes another try.  Lots of good test data for beckett rocket tek377a.


merbanan:
Looks good to me.

PR #139, opened 2017-01-21
In the process of collecting signals for all my devices

klattimer:
Working on lightwave rf and some generic controllers among other things. 

I've noticed that I can't get FSK to detect properly at 868Mhz btw, will investigate further.


merbanan:
Try to increase the sample rate.

PR #142, opened 2017-02-01
Add Signal file for BTHGN129

sbsrouteur:


sbsrouteur:
Looks like git is not for me. Can you just get the sample and trash the rest (and the PR as well)?

merbanan:
:) sure, I'll fix it later.

zuckschwerdt:
Looks like the samples are already merged to tests/oregon_scientific/thn129 now.

PR #148, opened 2017-03-27
Recordings for WG_PB12v1

ciaranmooney:
A set of recordings for the WG_PB12v1 temperature sensor. These demonstrate the temperature value changing and the ID value changing.

Also included are photographs of the base station, sensor, and sensor PCB.


ciaranmooney:
Apologies, pull request closed because I had forgot to tidy up commits before submitting.

PR #149, opened 2017-03-27
Recordings for WG_PB12v1

ciaranmooney:
Files necessary for testing WG_PB12v1 sensor. Recordings are of the temperature change and id changes. Note, not all the values of the packet have been figured out. The recordings are only for those that I'm confident of. The README.md describes the files included.


rtl_433_test.sh
```diff
@@ -1,7 +1,7 @@
 #!/bin/bash
 
 ## Your git project directory
-PRJDIR=~/Projects
+PRJDIR=~/Code
```
merbanan: Looks like a stray commit. Please remove.

ciaranmooney:
Correct.

PR #155, opened 2017-05-04
Support for temp-sensor from AURIOL 4-LD4437-1

hoenes:
I added sensor data (and images) and also identified most parts of the (decoded) signal, so I hope you can easily add support for this sensor - thanks so far, great work!


merbanan:
Something is missing from the patch. It is only 1 byte large.

hoenes:
Seems I am not really know how to GitHub.
I uploaded files again hopefully this time at the right location ;-)

PR #199, opened 2017-12-08
New Files and demod tool

fraschizzato:


zuckschwerdt:
can you add a file `samplerate` with just the sample rate number (e.g. `2000000`) to each folder where the sample rate is not 250k?

PR #201, opened 2017-12-17
Add honeywell doorbell test samples

Jc2k:


Jc2k:
See https://github.com/merbanan/rtl_433/issues/652

PR #205, opened 2017-12-20
New Data, New Demod/Decode Tool

fraschizzato:
Actually I need help on make device in rtl_433, 
modulation --> FSK_PULSE_PCM
short limit --> 26
long limit --> 26
reset limit --> 150

With the baud rate found (38400) i get a Samples per Symbol @ 200k (used in gnuradio clock recovery) of 5.2
So, if short limit is referred @ 1M sample rate i've set it @26

Now, I've uncommented first debug lines in new_template.c , but also using -D -D i didn't get anything.


klohner:
These sample files seem to be recorded with too high of a gain setting. It's difficult to recover the data.

I took a look at several of these files in Universal Radio Hacker and using tight bandpass filters, I could make out FSK data.  No idea what any of this data means, though.

Here's what I found in the files I looked at.  Hope this helps.

```
01/gfile041.cu8
1101010101010101010101010101010101101001110010001110100111001000100011100001100000000000000000000010101111100000000000000000000000101100110100000000000011101111000000000000000000000000000000000000000000000000011111010000110100011000001000111010000000000001000000001010000011010011001110000001110101011000101110100

01/gfile064.cu8
1101010101010101010101010101010101101001110010001110100111001000100011100001100000000000000000000010101111100000000000000000000000101100110100000000000011101111000000000000000000000000000000000000000000000000100111100000110100011000001000111010000000000001000000001010000011011011000011110001110110001101111101000

02/gfile007.cu8
110101010101010101010101010101010110100111001000111010011100100010001110000110000001000100

02/gfile008.cu8
1101010101010101010101010101010101101001110010001110100111001000100011100101000000000000000000000010100111100000000000000000000000101100110100000000011010001100000100111000111100000101100010001000001000010100000001101000110000000110100011110111111111111111100000001000000000000101000000000000000000010110010001011

02/gfile009.cu8
1101010101010101010101010101010101101001110010001110100111001000100011100001100000000000000000000010010111100000000000000000000000101100110100000000000011100000000000000000000000000000000000000000000000000000010011011000011010001100000100111101000000000000100000000101000000000110100000110000111011000011010100111

03/gfile762.data-20171207_132407.cu8
1101010101010101010101010101010101101001110010001110100111001000100011100101000000000000000000000010100111100000000000000000000000101100110100000000011010001100000100110000001110000110000010001000001000010100000001101000110000000110100011110111111111111111100000001000000001101111000000000000000000111010111001011

03/gfile763.data-20171207_132408.cu8
1101010101010101010101010101010101101001110010001110100111001000100011100001100000000000000000000010010111100000000000000000000000101100110100000000000
```

fraschizzato:
sorry, i didn't remember about that post...
anyway, actually i demod the files with the attached gnuradio flow (some var must be tuned each some time)
demod Decode grc
there are three different transmission every day
[one_day_data.zip](https://github.com/merbanan/rtl_433_tests/files/3154246/one_day_data.zip)
two of them are data from the tank, the third is a full timestamp.
The first type has that structure (in hex):
- a0 00 00 00 
- 57 or 4b (data type)
- c0000059a0 (device serial)
- first two hex digit (ie 05 f1) data from first input, seems to be liters in tank 1501 in decimal (about 85%)
- 00 00 00 00 - probably data from the other (unconnected) input
- two hex digits (ie 49 00) - only moves when tank refill occurs (usually 00 00)
- probably crc
- three hex timestamp --> converted to dec it's HH MM SS
- probably another crc
- other 7 digits (ie 00 00 00 00 00 00 00) unknown but probably includes another crc and battery level

The second type has:
- same as before
- 53
- device serial
- other data that includes: day, month, year, day-of-week (0-6), hh, mm, ss (repeated two times, so probably is a pairing between the two device and/or the time setting for next day transmission )  

010101 etc is preamble

data


Some steps were discussed here: https://groups.google.com/forum/#!searchin/rtl_433/urmet%7Csort:date/rtl_433/mQpw6MbRfXU/5YkcqV00AgAJ

merbanan:
The new prototype FSK demod algoritm was able to do pulse recovery of one tested sample file.

fraschizzato:
I will take a look on that, i didn't see how new fsk works but in mine demod tool I have to change center frequency for each file type. Usually the timestamp file has a center frequency higher than the other two, but sometimes are three different frequency. In gnuradio flow I made this adjustment in the FIR filter, as you can see the "center frequency" var is at -25k.  By time i've used values from -40k to 60k. 
Didn't know if it depends from signal or from my capture device but about frequency i've seen:
- Different frequency for different data packet (adjusted in FIR filter)
- Little frequency difference by time (adjusted in FIR filter)
- "Bigger" frequency difference by time (changed grab frequency on rtl_433)

merbanan:
The new fsk demod is vaporware living on my computer. I'll try to clean it up so you can actually use it for offline cases. It is written in octave to ease the prototyping. Anyway so far it has been performing very well, it has quite a few to many conditionals for my liking though.

PR #210, opened 2018-02-15
Add in captures of Byron BY101 doorbell

mzealey:
`-a` and `-A` return sensible looking values, but I can't seem to figure out how to get that working within a driver - any advice appreciated.


zuckschwerdt:
see https://github.com/merbanan/rtl_433/issues/682

PR #211, opened 2018-02-17
Add File for quhwa doorbell

peteakalad:
Add File for quhwa doorbell (3),  not decoding for me


zuckschwerdt:
Can you remove the "used to work" "broken" "not working" comments? we'll fix it and there is no reason to keep that note ;)

PR #212, opened 2018-02-17
Add test data for Conrad Pool Thermometer

Philippovitsch:


zuckschwerdt:
Are the samples taken with default settings, i.e. 433.92M and 205k? Can you rename them to e.g. t244c_433.92M_250k.cu8 so there is no ambiguity?

Philippovitsch:
Yes, they are taken with default settings, changed the file names, thanks.

Philippovitsch:
Got an idea of extracting the temperature, attached a file with my thoughts.
The rest of the message are the same bits everytime, wheter I re-insert the batteries or pressing the reset button, except the last 4 bits.

[Decoding.pdf](https://github.com/merbanan/rtl_433_tests/files/1737369/Decoding.pdf)

zuckschwerdt:
please use an issue at rtl_433 to discuss.

PR #214, opened 2018-02-28
Update README.md

firefighter76:
Add code with battery low signal.


zuckschwerdt:
this is already merged and I commited the update on top now.

PR #215, opened 2018-03-04
Added samples for intertechno ITT-1500

h3po:
There should be many more samples for intertechno devices in the future, so I put the readme into the subfolder along with the samples. How are test fixtures going to be defined?


h3po:
see **https://github.com/merbanan/rtl_433/pull/694**

PR #220, opened 2018-05-24
Add Oregon Scientific UVR128 UV Sensor samples

ovrheat:
Add Oregon Scientific UVR128 UV Sensor samples.


zuckschwerdt:
Are these recorded on the default sample rate and center frequency?

zuckschwerdt:
All is well. I'd just like to rename the files for future reference.

ovrheat:
Ah, ok :) These are several recordings, picked up in many different times anyway... I renamed all in that way just to have some kind of order.

zuckschwerdt:
The params are not terribly important, just nice to have. See https://github.com/merbanan/rtl_433_tests/commit/0382340

ovrheat:
That's right. I'll keep it in mind for the next time ;)

PR #223, opened 2018-06-30
unknown chinese remote with silk touch

Grf2017:
Data for an unknown black chinese remote with silk touch, two buttons mode and on / off.


zuckschwerdt:
Push-button (doorbell) style devices don't need a custom decoder. Something like
```
rtl_433 -q -F json -R 0 -X 'black:OOK_PWM:264:512:1200:700' -r g001_433.92M_250k.cu8
```
should work.

PR #226, opened 2018-07-06
Add generic 4ch black remote 433MHz

mcauser:


zuckschwerdt:
Looking at the first sample this is an interesting test case as the sample is so distorted. AM is rather low with high spikes. It's simple OOK with fixed period of 1300 µs with long pulse at 1000 µs, short pulse at 350 µs. The first sample should read `1000010000011001000001000` / `0x841904` (using long for 1, i.e. inverted). 25 bits as expected for this kind of device.

zuckschwerdt:
Forcing a very low trigger level will also show it in the am analyzer: `rtl_433 -q -l 1000 -a -r g001_433.92M_250k.cu8`

PR #230, opened 2018-07-28
Add Hyundai WS SENZOR

Koumes:
Add Hyundai WS SENZOR description and testing data.


Koumes:
Yes . It is so.

rtl_433 -a -t

PR #232, opened 2018-08-11
Create readme.md

RadioBon:


zuckschwerdt:
See https://github.com/merbanan/rtl_433/blob/master/decoders.cfg
I.e. thi should work
rtl_433 -q -F json -R 0 -X 'Byron_BY_Doorbell:OOK_PWM:500:1000:3300:1200,match={2}0x3,repeats>=4,bits=21'
There are some changes comming, you should be able to use that config directly soon.

PR #234, opened 2018-09-13
New device recording for fineoffset_wh65b

StephenR0:


zuckschwerdt:
Fully supported with the WH25 /WH24 code now.

StephenR0:
It works just fine on the my test sensor array.  I can't get anything from my outdoor array, but I suspect my antenna isn't good enough.  I'll keep testing.  Thanks.

StephenR0:
It works on my outdoor array as well.  I'm having some signal strength issues.  I brought it inside to change the batteries and reset the array.  While I had it inside, I tested it and rtl_433 had no trouble getting the data.  Thanks.

zuckschwerdt:
Get an antenna tuned to the frequency (915 here, usually 433 or 866). Universal and mismatched antenna have a noticeable loss (DVB-T antenna would be something like  474-858).
Never use the RTL-SDR without an USB extension cord. There is much interference at the computer.

StephenR0:
Thank you for your advice about antennas.  I'm still learning about sdr and I'd been suspecting that a better antenna would be helpful.  But the problem that I've been having is between the sensor and the console of the weather station.  Since it's only about 40 feet between them, I would have thought it would be more reliable.  But I'm still trying to figure out where the problem is.  It seems to come and go.  There are a couple of walls for the signal to go through, though.  Ultimately, I'd like to build up a Raspberry Pi to receive the signal for Weewx.  But one thing at a time.  :-)

zuckschwerdt:
Maybe try CubicSDR or Gqrx to visually inspect the live signal for interference and if the center frequency is ok. You don't want the signal right on the center frequency and also not to far out. Pick a value that covers all your sensors. Default sample rate of 250k gives center +/-250k.

PR #238, opened 2018-10-16
Update on TFA 30.3211.02

ionum:
Decoded all values execpt battery low....


ionum:
@merbanan : should i create a pull-request for merbanan/rtl_433 ?

PR #244, opened 2018-11-06
Add tests for hideki/cresta temperature

Mindavi:
related to https://github.com/merbanan/rtl_433/pull/858


Mindavi:
The actual module looks like https://jeelabs.org/2010/04/16/cresta-sensor/, but I don't have own photos as of now.

PR #245, opened 2018-11-20
BMW/Continental TPMS Sensor Recordings + Description

FizzyWhiz:


zuckschwerdt:
Thanks! The `01`… directories should be used for different devices of the (supposed) same kind, say rebranded devices or from different contributors. I'm going to shuffle all samples here into the `01` dir :)

zuckschwerdt:
Using `rtl_433 -q -F json -X continental:FSK_PCM:50:50:150 -r file.cu8` should show the bits for most TPMS. This is the usual FSK with 50µs half-bit width here, but… GFSK is used and that doesn't play well with the current demodulator :/ Supporting GFSK is a major overhaul to the demod but high on my list.

zuckschwerdt:
I might get around to test some ideas for GFSK in the comming weeks. Otherwise gnuradio works well.

zuckschwerdt:
If you drop PD_MIN_PULSE_SAMPLES in [pulse_detect.h](https://github.com/merbanan/rtl_433/blob/master/include/pulse_detect.h#L19) to 4 you can get the data with:

    rtl_433 -q -F json -R 0 -X 'continental_tpms:FSK_PCM:26:26:150' -r file.cu8

The data is Manchester encoded (that's why you only see 5, 6, 9, and a), leading `a`s are a warm-up preamble.

E.g. your first sample file has:
```
{214}aaaaaa5995555a9a9569a59665665559659a6595555696566a5afc
{206}aaaa5995555a9a9569a59665665559659a6595555696595559f0
{214}aaaaaa5995555a9a9569a59665665559659a65955556965a59a6fc
```
align on the first non-preamble bit you get
```
5995555a9a9569a59665665559659a6595555696566a5afc
5995555a9a9569a59665665559659a6595555696595559f0
5995555a9a9569a59665665559659a65955556965a59a6fc
```
which looks plausible an decoded is (not sure wether MC or DMC):
```
MC d7fc47936bafdb4b7fe6e8c
11010111111111000100011110010011011010111010111111011011010010110111111111100110111010001100

DMC bc02645ade7836eec0159ca
10111100000000100110010001011010110111100111100000110110111011101100000000010101100111001010
```
Maybe use [this BitBench](http://triq.net/bitbench?c=d7fc47936bafdb4b7fe6e8c&c=bc02645ade7836eec0159ca&f=8h%20) and shift the bits and see if you spot the serial number there.

zuckschwerdt:
Thinking about it, readings should work if you run at a higher sample rate. Use `-s 1024k` and the `-X ...` should work. You need new samples for that though. If you acquire samples at 1024k please upload a few as `02`.

FizzyWhiz:
Sure, I can grab some new samples. How does -s 1024k alter things?

zuckschwerdt:
Option `-s` sets the receiver sample rate. Default is 250k, the next faster setting for RTL-SDR is 1024k, roughly 4x more samples. The default demod limit is to accept a "bit" only if it is stable for more than 10 samples. Your (half-)bit length is 25 µs, which is *very* short, only half of what other TPMS use (which are the fastest protocol currently in rtl_433). 25 µs at 250k is only 6 samples, at 1024k it would be 25 samples :)

FizzyWhiz:
Finally got around to trying to get the 1024k sample. However running -s 1024k switch tuned to 433mhz the output floods the screen, about 20 .cu8 in 4 seconds, same issue with -X 'continental_tpms:FSK_PCM:26:26:150' . Also have a new model of Toyota TPMS sensor that doesn't current register automatically on RTL, shouldn't hurt to gather this with  -s 1024k too as well?

zuckschwerdt:
New sensor samples with a description of the data are always useful to enhance support and build a regression test. Can say what's wrong with the -s 1024k, really should explode like that. I'll check a few things if something obvious is wrong.

PR #248, opened 2018-12-06
Honeywell ActivLink Samples Submission and Documentation

klohner:
Submitting 4 variations of a signal with the same device ID.  Also submitting documentation on this signal as I've been able to understand it based on a very limited sampling of signals.


zuckschwerdt:
I have numbered the sample (g00x_). The current parsers can't detect sample info at the start of the name.

PR #249, opened 2018-12-17
Add Digitech XC0346 samples (rebadged Fine Offset WH1050)

projectgus:
Please let me know if you need any more details for this. :)

Patch with rtl_433 support: https://github.com/merbanan/rtl_433/pull/922


zuckschwerdt:
Is this similar to the device in fineoffset_wh1080/04? It hints to XC0346 there.

projectgus:
I think it goes like this:

WH1080 == XC0348
WH1050 == XC0346

WH1050 and WH1080 are similar but not the same. They currently have separate decoders in rtl_433, although it looks like there are some similarities in the packet format.

OEM seems to be:
* http://www.foshk.com/Weather_Professional/WH1070.html (WH1050 is the version of WH1070 without RCC)
* http://www.foshk.com/Weather_Professional/WH1080.html

PR #256, opened 2018-12-31
Add one reccord + picture + notes in readme for Zeepin TPMS

enavarro222:


klohner:
I took a look through these samples.  In several, there seems to be overlapping signals with some other device.  Also, perhaps they were recorded at too high of a gain setting making decoding somewhat difficult.

However, for example, here's a representation of what's in `02/g059_433.92M_1024k.cu8`.

zeepin_g059

With a bit length of 54 µs, it looks like two repetitions of `{160}0xaaaaaaaa5a9a5a9a6aa99a5555aa695a9aa55a9a` which seems to be some kind of Manchester encoding.  I'll assume it's Manchester (G.E. Thomas) but it might very well be Manchester (IEEE 802.3) or something else.

Assume the first 32 bits of that is a preamble, we're left with 128 bits of encoded data.  After Manchester (G.E. Thomas) decoding (with [BitBench](http://triq.net/bitbench?c=0xaaaaaaaa5a9a5a9a6aa99a5555aa695a9aa55a9a&f=PREAMBLE%3A16h%20DATA%3A64h&d=MC)), we get 8 bytes of data: `0x3b3b7eb00f63bc3b`.  This aligns with the reading `rtl_433` finds.

`rtl_433 -s 1024k -X 'n=zeepin,m=OOK_MC_ZEROBIT,s=50,l=50,r=1000' -r g059_433.92M_1024k.cu8`
`    {84}0ffff3b3b7eb00f63bc3b,{80}00003b3b7eb00f63bc3b,{15}0000`

From the provided reading on the receiver for this sample, the data should align to a temperature of "19°C" and a pressure reading of "0.1 bar"

|.cu8|Manchester (G.E. Thomas) Decoded Data|Temp|Pressure
|----|------------|----|--------
|g010|53536dc8f77ba453|11°C|0.1bar
|g059|3b3b7eb00f63bc3b|19°C|0.1bar
|g073|3b3b7eb00f63bc3b|19°C|0.1bar
|g086|434375b8076bb443|20°C|0.1bar
|g100|434375b8076bb443|20°C|0.1bar
|g152|5f5f64d4eb87995f| 9°C|0.1bar
|g177|1f1faa942b47d91f| 3°C|0.1bar

For each sample, the first two and last bytes are identical.  Perhaps this denotes some kind of marker.  To reliably find the end of the preamble, though, I'd guess this byte would never be greater than 0x7f, so it's maybe not a CRC.  And it varies for each different reading, so it probably means something other than some kind of data prefix/suffix.

We're then left with 40 bits of data (see [BitBench](http://triq.net/bitbench?c=0x53536dc8f77ba453%20%5Bg010%2C%2011%C2%B0C%2C%200.1bar%5D&c=0x3b3b7eb00f63bc3b%20%5Bg059%2C%2019%C2%B0C%2C%200.1bar%5D&c=0x3b3b7eb00f63bc3b%20%5Bg073%2C%2019%C2%B0C%2C%200.1bar%5D&c=0x434375b8076bb443%20%5Bg086%2C%2020%C2%B0C%2C%200.1bar%5D&c=0x434375b8076bb443%20%5Bg100%2C%2020%C2%B0C%2C%200.1bar%5D&c=0x5f5f64d4eb87995f%20%5Bg152%2C%20%209%C2%B0C%2C%200.1bar%5D&c=0x1f1faa942b47d91f%20%5Bg177%2C%20%203%C2%B0C%2C%200.1bar%5D&c=&c=&c=%5B%5D&f=MARKER%3A8h%208h%20DATA%3A40b%20MARKER%3A8h&a=Preamble&m=)) which I'm assuming should contain an ID, temperature, pressure, alert flags, etc.  I don't see an obvious way to attack that problem from the data on hand. None of 

- Are these samples all from one of the devices or from all 4?  This needs to be identified in each sample.
- For each device, we should probably get several identified readings where just one of these variables changes slightly.  For example, (temperature changes by 0.1 degree, pressure constant), (pressure increased by 0.1 bar, temperature constant), (pressure decreased quickly which should trigger an alert flag), etc.

@enavarro222, can you provide such samples?

PR #258, opened 2019-01-06
Add XH300/XT300 description + record + test json

enavarro222:


Mindavi:
Do you have any examples of the sensor set to another address (the linked website says there are 3 different addresses)? Might be useful for testing if the channel decoder works correctly.

Mindavi:
Looks good to me now

zuckschwerdt:
This has JSON files, best to merge at the same time as the device. The CI gets upset otherwise :p
The "model" name really should be "Opus-XT300".

PR #266, opened 2019-02-26
Add samples for KNX RF

holmie:
PR contains recorded (rtl_433 .... -S all) samples for KNX RF.
Please advise if there are better ways of recording these.


zuckschwerdt:
That's the preferred way to record samples :) Did you check the [spectrogram](http://triq.net/iqs) to verify the samples show what you wanted to capture?

holmie:
Checked it now, all though I'm not very proficient in interpreting the things I am looking at.
I guess I need some help in actually understanding/decoding this, if it's even possible.

zuckschwerdt:
Looking at your first sample all is well. Clear band and two sensors transmission that are not clipped (don't run into the left or right end of the "picture") and not overlayed be some other signal.

holmie:
Great, I can just hope someone sees this a challenge and can provide some help or pointers on decoding it. :-)

merbanan:
Ok, so when I add this (the events condition is never triggered):

```
+            int newevents = 0;
+            if (device->decode_fn) {
+                events += device->decode_fn(device, &bits);
+            }
+            // Debug printout
+            if (!device->decode_fn || (device->verbose && events > 0)) {
+                fprintf(stderr, "pulse_demod_manchester_zerobit(): %s \n", device->name);
+                bitbuffer_print(&bits);
+            }
```
To the pulse_demod_manchester_zerobit() code I get output like this:

time      : @0.227374s
model     : KNX-RF       count     : 9             num_rows  : 9             rows      : 
len       : 1            data      : 0, 
len       : 1            data      : 0, 
len       : 1            data      : 0, 
len       : 1            data      : 0, 
len       : 1            data      : 0, 
len       : 1            data      : 0, 
len       : 175          data      : 6bb00fcfff6f9bffe6b1ad1fffa00fffd2fff7ea6ac0, 
len       : 2            data      : 0, 
len       : 1            data      : 0


The top 5 bits matches the sync word from the documentation. So the manchester decoder needs to changes and the fsk demodulator/bit slicer needs to be fixed also as it misses half of the signals from the supplied samples.

zuckschwerdt:
Might be worthwhile to check what the decoder sees with the Sigrok pulseview. The demod is not well suited for GFSK and alike.

holmie:
Let me know if you want other (or better samples).
I have 6 sender devices and about the same amount of receiver (switches).

One sender does some kind of group send to turn on/off all receivers.

PR #268, opened 2019-03-20
honeywell 5890PI

fanlessfan:
honeywell 5890PI motion sensor
http://library.ademconet.com/MWT/fs2/5890PI/5890PI-Data-Sheet.PDF


fanlessfan:
Finally, I attached the files.
[5890PI.zip](https://github.com/merbanan/rtl_433_tests/files/2988159/5890PI.zip)

zuckschwerdt:
works, I'll test and merge this probably tomorrow.

fanlessfan:
Thank you @zuckschwerdt

merbanan:
The samples are all over the place. You are picking up other random signals. Also what ISM-band zone are you in? 315 or 433? We need more consistent sample files. One way to achieve that is to remove the antenna of the rtl-sdr receiver and the place the transmitter in the proximity and trigger a transmission. Or use a manual gain setting (with low enough gain to only pick up the relevant signal).

Right now there looks to be 3-6 different signals and documentation online indicate that the transmitter should use a protocol that we already have support for. So we really need to sort out the red herrings so we don't focus on the wrong signal.

fanlessfan:
the sensor is 345MHz. Please see the FCC ID https://fccid.io/CFS8DL5890-1. Here is the new files and hope it has good signal. I removed the antenna. I also record the signal using URH. https://github.com/jopohl/urh

[g014_345M_250k.zip](https://github.com/merbanan/rtl_433_tests/files/2998089/g014_345M_250k.zip)
[RTL-SDR-345MHz-1MSps-1MHz.zip](https://github.com/merbanan/rtl_433_tests/files/2998090/RTL-SDR-345MHz-1MSps-1MHz.zip)

fanlessfan:
here is the screenshot of URH
image

merbanan:
Ok, the sample signals look good. Now the bad news. They are PSK modulated and rtl_433 has no PSK demodulator. But now that we have PSK sample signals we have a reason to write the code. But as usual no guarantees when or even if that code will be written.

fanlessfan:
Thanks @merbanan Hope you can write the code eventually. let me know if you need more sample signals.

klohner:
Just out of curiosity, I tried poking around at these samples and made some progress.

Although they seem to be PSK at first glance, I couldn't make sense of a PSK demodulation -- each demodulated repetition was different.  But, if ignoring phase and just trying to demodulate as ASK (not OOK, it's actually modulating between two different amplitudes), suddenly there are repeated data messages.  Going through all the samples, there's only 3 unique ASK data messages:

|ASK Decoded|Seen in sample file
|-----------|------------------
|0x00555555566aa5a5a99a6696aa6aa66656|[354MHz, g002, g003]
|0x00555555566aa5a5a99a66a6aaaaa5a659|[345MHz, g003, g004, g005, g009]
|0x00555555566aa5a5a99a6666aa6aaaa656|[g004, g005, g007, g008]

These seem to be some kind of Manchester encoding.  On a whim, I tried feeding these samples into rtl_433 with a manual level specified, and they were recognized.

`rtl_433 -r g002_344.975M_250k.cu8 -l 15000`
`rtl_433 -r g003_344.975M_250k.cu8 -l 15000`
`rtl_433 -r g004_344.975M_250k.cu8 -l 15000`
`rtl_433 -r g005_344.975M_250k.cu8 -l 15000`
`rtl_433 -r g007_344.975M_250k.cu8 -l 15000`
`rtl_433 -r g008_344.975M_250k.cu8 -l 15000`
`rtl_433 -r g009_344.975M_250k.cu8 -l 15000`

These all get recognized as various events on a "Honeywell Door/Window Sensor" with ID "3314a".  That seems to make sense to me.
___

Additionally, there seems to be another type of pure ASK/OOK signal in other files:

|ASK/OOK Decoded|Seen in sample file
|---------------|------------------
|0x55555556a95aaa6aaaaaaa5aaa6aa66969a5aaaa66aa66aa66aa5a659a66|[g001]
|0x55555556a95aaa6aa9aaa95aaa6a9a6a59a5aaaa66aa66aa66aa5a659a66|[g004]
|0x55555556a95aaa6aa6aaaa5aaa6a6665a9aaaaaa66aa66aa66aa5a659a66|[g006]

Again, these look like some kind of prefix and Manchester encoding.

|Manchester (G.E. Thomas)|Seen in sample file
|------------------------|------------
|e3f7fff3f7d66cff5f5f5f34b5|[g001]
|e3f7efe3f7b72cff5f5f5f34b5|[g004]
|e3f7dff3f754efff5f5f5f34b5|[g006]

Setting levels manually and using -vv, I could get rtl_433 to find the data, but it doesn't seem to want to decode them.  Any idea what is broadcasting these signals?

klohner:
Also, this was using the samples in the first `5890PI.zip` file, not the ones in the `g014_345M_250k.zip` file.  When I look at the samples in `g014_345M_250k.zip`, I get a sense that the gain was too high when capturing and the different ASK levels in the signal are blown out. This leaves only the misleading phase info, making decoding using ASK in `rtl_433` impossible.

fanlessfan:
Hi, @klohner Could you give the instruction how to record the proper signal? thx

merbanan:
I think I have to retract the PSK decoding statement. It is most likely ASK but not OOK coded. With offset tuning it should then be possible to decode the signal to bits.

klohner:
You could try forcing a lower gain level in `rtl_433` like this:

`rtl_433 -f 345M -R 0 -g 20 -S all`

Try lower or higher -g levels.  If it's too low, `rtl_433` won't find a signal to save.

If you bring the signals into URH, you'll know the gain is too high if the top of the signal is cut off in the "Analog" Signal view.  You'll want to see a rough edge along the top.

klohner:
These `.cu8` files look very nice.  `rtl_433` even pulls out one of the repetitions of data from g001, g002, and g008 with the default auto level.

My first guess was `rtl_433` is not auto-adjusting its bit level threshold between the lower and higher ASK levels fast enough to pick up the data in these samples reliably.

A manual `-l` level setting of somewhere between 6700 and 14900 gets all the data messages in these samples.  10000 works very well.

`rtl_433 -r g001_345M_250k.cu8 -l 10000`

Trying to test my hypothesis, I turned on Analyze Mode both without and with the manual level setting.

`rtl_433 -r g001_345M_250k.cu8 -a`
and
`rtl_433 -r g001_345M_250k.cu8 -a -l 10000`

Surprisingly, in the first of those two commands, `rtl_433` with the default auto level seems to pick up all the data.  I wasn't expecting that.  But for some unknown reason, the Honeywell decoder doesn't seem to decode most of it and I'm not sure why.

In the second command, where `-l 10000` is manually specified, `rtl_433` seems to find the very same data, but this time the Honeywell decoder does a thorough job of finding all the data.

merbanan:
The -a analyzer does not use the same code path as the -A analyzer. Thus the output differs.

Mindavi:
I've added one of the recordings that generates an output from the honeywell-security decoder. I think support was added / improved in https://github.com/merbanan/rtl_433/pull/747.

PR #270, opened 2019-03-27
Add files via upload

ukshady:
Sample files for Oregon Scientific WGR800X Wind Sensor


zuckschwerdt:
The BitBench is about a TPMS? Perhaps wrong link?

PR #272, opened 2019-03-28
Additional sample files for Neptune R900.

bondiblueos9:


zuckschwerdt:
Oh, that's too much files to include. We need to keep an eye on repo size. Select ~5 interesting files.

bondiblueos9:
I'll pick 5 with differing values and write up the correct decoding for them.  What is the best sample rate for you to work with?  I cannot capture anything that makes sense with lower than 1000k.

zuckschwerdt:
Your center frequency is way off. Drop a sample in http://triq.net/iqs to see. Try to get the center frequency close to the signal frequency but about 20k off. Dead center isn't reliable.
I'll undo the other PR. Didn't check the file beforehand :/ Ok to reuse this PR and just sort things out with more commits though.

zuckschwerdt:
250k should be enough then.

bondiblueos9:
I was able to get some samples at 250k.  Unfortunately, I was only able to get them at about 50k off the center frequency, not 20k.  The device does not use a consistent frequency.  I added two samples each at three new consumption values, as well as an additional 1000k sample for one of them but closer to center than the first.  I visually inspected the signals, included the expected payload in the readme, and confirmed that they calculate to the correct values.

zuckschwerdt:
Deviation up to ~100k is fine at 250k rate. Just don't tune dead center ;)

PR #273, opened 2019-03-29
Update BTHGN129.json

jamaron:
JSON updated after pressure value calculation fix


zuckschwerdt:
Can you pull /rebase? This is already in.

PR #274, opened 2019-04-08
Garage door remote

Mindavi:
Hi,

I've recently taken some recordings of our garage door remote and was wondering if I could get some help decoding these. I suspect a rolling code is involved. The samples are recorded in order while pressing one of the buttons.

I've described a decoder in the README, which seems to decode the files. However, different files will have a different payload length, so I'm unsure if this is really correct.

Besides that, I needed to cut off the start (which seemed to contain no signal) off the recorded samples. Only after doing that a signal was actually seen and decoded by rtl_433.

[The edited files](https://github.com/merbanan/rtl_433_tests/files/3056115/edit-signals.zip)

I'm fairly certain it is _some kind of_ FSK signal, as the examples on wikipedia look a lot like the signal attached.

Could anyone help me out with decoding this signal, if possible?

There are some things that need to be fixed before this can be merged. I think decoding the signal first is the best way to go about this.

TODO:

- [x] picture of control unit
- [ ] picture of remote control
- [ ] description of signal

Some general notes/findings:

- The files must be chopped up before decoding correctly, this points me to a possible issue in the decoder.
- The suggested decoder is OOK, even though it seems to detect FSK after fixing up the files. This seems incorrect to me.
- After testing the new FSK decoder (minmax decoder), it seems like this might be a KeeLoq type message. Ergo, partially encoded, partially plain data. Seems promising.

```
Detected FSK package	@0.004100s
...
Frequency offsets [F1, F2]:   24794,   4051	(+94.6 kHz, +15.5 kHz)
Guessing modulation: Pulse Width Modulation with sync/delimiter
Attempting demodulation... short_width: 392, long_width: 780, reset_limit: 15884, sync_width: 592
Use a flex decoder with -X 'n=name,m=OOK_PWM,s=392,l=780,r=15884,g=0,t=0,y=592'

```

New decoder:
```
decoder {
  name=remote,
  modulation=FSK_PWM,
  short=392,
  long=780,
  reset=15900,
  tolerance=150,
  gap=1000,
  sync=0
}
```
@Mindavi update with the files on linux pc.


zuckschwerdt:
The samples look great. The FSK signal is not DC free, the lower frequency is by far dominant. That's a problem for the current demod. @merbanan is working on a improved decoder.

Mindavi:
According to this website: https://www.weatherstop.fr/sommer-aperto-868-l.html

> Extremely safe against unauthorised entry! The “Rolling Code System” changes automatically each time the operator is activated.

If I want to use a remote, I also need to pair it first.

PR #275, opened 2019-04-17
TFA-KlimaLogg-30.3180.IT TestData (merbanan#262) ver2

pjenglund:
I collected some samples from at least four TFA 30.3180.IT sensors. 

I have noticed some other signals on the 868 MHz frequency but I hope they are not interfering.


zuckschwerdt:
That looks like a good collection of signals! But we need to filter those down to the most interesting ones and likely downconvert to 250k. This repo musn't grow in size too fast.

pjenglund:
Ok, I'll make those changes!

zuckschwerdt:
When capturing or resampling make sure the actual signal isn't exactly centered, the DC component is often a problem. The signal should be e.g. 20 kHz off center.
You can quickly look at the signals and often spot a few parameters with http://triq.net/iqs

pjenglund:
What would be the easiest way to delete the original [cu8] files and replace them with new ones (or just keep a few of the most interesting ones)? Should I close this pull request and do a new one? I tried to mess with the original pull request but since I'm quite new to Git, I'm not sure how it worked out...

zuckschwerdt:
You can just add more changes on top as you did. It already looks good.

PR #276, opened 2019-04-21
Create WeatherDuino AQI module

41south:
To add decoding of the WeatherDuino AQI module into Weewx via rtl_433 and the Weewx-rtl plugin


zuckschwerdt:
Note that it actually emulates the Davis Vantage Pro2 communication protocol, should work for both then.

41south:
Except that the Davis protocol has no room (that I am aware of) for AQI so
I was going to use the rtl-sdr driver to grab it. In fact I may now go down
another route all together and grab the UDP broadcast from WeatherDuino
which also contains the AQI data

zuckschwerdt:
Do you have a link with the protocol description? I only found things related to the USB or network data from the console.

41south:
No longer required :)

PR #277, opened 2019-05-06
Add Companion WTR001 Temperature Sensor Test Data

klohner:
Here is test data for the Companion WTR001 Temperature Sensor provided by "B Rabbit" to the rtl_433 discussion list.  I've added a description of the signal format.


zuckschwerdt:
Looks good, thanks! Can you resize/resample the jpgs to about 100K size (there doesn't seem to be too much detail in there)? As always we need to keep an eye on overall size.

klohner:
Thanks.  I cleaned up a couple small things after your merge if you'd like to remerge.

zuckschwerdt:
The last commit is 26 minutes ago "Delete WTR001_user_manual.pdf", I merged after that. I guess we can reset and remerge if there is something too change after that?

klohner:
Nothing major - just renamed a couple images to be consistent and replaced "us" with "μs" because it bothered me.  I see that change in my branch, but I guess it doesn't get picked up here after you merged.

zuckschwerdt:
Ok, maybe just open another PR from the same branch then?

PR #286, opened 2019-06-05
Improve support for cheap EV1527 based PIR sensor

curlyel:
EV1527 based devices are supported (at least?) in the Generic remote [30] and Wireless Smoke and Heat Detector GS 558 protocol [86]

A EV1527 based PIR from brand SGOOWAY model PR2 is detected as

```
time      : 2019-06-05 14:06:02
model     : Smoke detector GS 558                  id        : 11148
unit      : 19           learn     : 0             Raw Code  : 357193
```

The reception is quite poor. Just one wall in between the sender and the rtl_sdr breaks the "connection". This has been tested with two different rtl-sdr sticks and two different antennas (one dedicated 433MHz antenna and one DVB-T antenna which came with one of the radio sticks).

A lot of other 433MHz sensors (Oregon, Alecto, Globaltronic) can be received from other floors (even from within a fridge downstairs ;-).

I'm going to provide new samples for it to help improving its reception/decoding. 

Could you check the samples and advise how to improve the decoding/reception (if it is possible at all ;-))

btw.: Similar thing has been dealed with at: https://github.com/merbanan/rtl_433/issues/663


curlyel:
Curious to hear, if the provided stuff is sufficient.
If not, please advise ;-)

zuckschwerdt:
Sorry, for the late reply. If you don't have the GS 558 you should disable that protocol. It's very likely to produce false positives with other EV1527.
The best and very flexible way to support your EV1527 device is to add a conf file with a flex spec, there are some examples for PIR and others.
The poor reception could be the timing not matching the GS 558. I'll take a peak at your sample files and suggest a flex spec.

curlyel:
Great - Thanks a lot for your help!

zuckschwerdt:
Timings like this should work:
```
rtl_433 -X 'n=sgooway,m=OOK_PWM,s=400,l=1200,g=1500,r=12000'
```

zuckschwerdt:
If you need help with the conf file open an issue on rtl_433. It would also be great to add your finished conf file to the examples, please PR :)

curlyel:
Much better now - not perfect yet but quite o.k. for such a cheap device ;-)
Let me check this in a less noisier environment later ...

curlyel:
> If you need help with the conf file open an issue on rtl_433. It would also be great to add your finished conf file to the examples, please PR :)

Does passing the config parameter via a config file having some advantage? Up to now I always passed the settings via the command line. My current tests regarding the EV1527 PIR for example using this cmd line:

```
rtl_433_cmd = "/usr/local/bin/rtl_433 -q -M protocol -R 12 -R 25 -R 79 -F json -X 'n=sgooway,m=OOK_PWM,s=400,l=1200,g=1500,r=12000' "
```
This is called from inside of my `rtl433-to-mqtt.py`. I've modified the json-to-mqtt conversion to fit my need and to ease the parsing/configuration in my smarthome hub (openHAB):

```
for line in iter(rtl433_proc.stdout.readline, '\n'):
        if "time" in line:
            mqttc.publish(MQTT_TOPIC, payload=line,qos=MQTT_QOS)
            protocol="P0"
            channel="C0"
            id="I0"
            data="D0"
            data_val=0
            data_cnt=0
            json_dict = json.loads(line)
            for item in json_dict:
                value = json_dict[item]
                if "protocol" in item:
                    protocol="P"+str(value)
                if "channel" in item:
                    channel="C"+str(value)
                if "rows" in item:
                    for row in json_dict['rows']:
                        if row['data'] == data_val:
                                data_val = row['data']
                                data_cnt += 1
                                data ="D"+str(data_val)
                        elif data_cnt < 5:
                                data_val = row['data']
                                data_cnt = 0
                    if data_cnt <= 5:
                        data="D0"

            subtopic =protocol+"/"+channel
            if protocol == "P0":
                subtopic +="/"+data

            if not(protocol == "P0" and data == "D0"):
                for item in json_dict:
                    value = json_dict[item]
                    if not ("brand" in item or "model" in item or "channel" in item or "protocol" in item or "rows" in item or "codes" in item):
                        mqttc.publish(MQTT_TOPIC+"/"+subtopic+"/"+item, payload=value,qos=MQTT_QOS)
```
This creates me subtopics like  `P<protocol>/C<channel>/<item>` 
_Note: `ID`  is not part of the topic to overcome the need to adopt the topics in case of a battery change which leads to new ID's of my sensors_
 e.g.:

```
sensors/rtl_433/P12/C3/battery OK
sensors/rtl_433/P12/C3/temperature_C 23.8
sensors/rtl_433/P12/C3/id 5
sensors/rtl_433/P12/C3/humidity 52
sensors/rtl_433/P12/C3/time 2019-06-14 18:19:59
```
Which can easily be consumed by openHAB's mqtt-binding. For the EV1527 I added the check, if the decoded data is the same at least five times per json. This should prevent false positives being published to the MQTT broker.

zuckschwerdt:
conf files are equivalent, but easier to add matches and alike and also document :)

The `-F mqtt` does much the same MQTT mapping if you use `usechannel=replaceid` and `devices=sensors/rtl_433`.
Use that with -M newmodel to get better (shorter, clean) protocol names.

curlyel:
> The `-F mqtt` does much the same MQTT mapping if you use `usechannel=replaceid`

Pretty cool! Never noticed these options, must be reasonable new - right? Thanks for that hint!

I've seen in the `rtl_433_example.conf` something like:
```
repeats>=4,bits=21,match={2}0x3
```
Is it something to filter out unwanted decodings and to tame the write load to output (e.g. to the mqtt broker)? If so, can you elaborate a bit on this or point me to some documentation?
Maybe this makes my Python-filter/conversion script superfluous ... ;-)

> conf files are equivalent, but easier to add matches and alike and also document :)

O.k. got it (and agree ;-) ) I'm going to provide my `rtl_433.conf` once its done for the PIR ...

zuckschwerdt:
Yes, native MQTT is very recent -- if there are requirements for a specific format to better fit consumers like home automation system please request that.

There isn't much documentation on the flex spec other than the `-X` help output and the conf samples.
It should be pretty much what you expect:
- `repeats>=4`: require at least four identical repeats
- `bits=21`: require exactly 21 bits
- `match={2}0x3`: require a two bit match (of `11`) somewhere

curlyel:
> if there are requirements for a specific format to better fit consumers like home automation system please request that

Well, it would be really great, if some "IoT standard convention" could be selected by a conf parameter.
At first: The "Homie convention" come to mind:

https://homieiot.github.io/

Because of the self-descriptive structure of it, a home automation system can automatically parse it and generate items/objects out of it. David Gräff, the maintainer of the (new) openHAB MQTT-binding has done a blog post about it:

https://www.openhab.org/blog/2018-12-16-mqtt-arrives-in-the-modern-openhab-2-x-architecture.html#mqtt-thing-auto-discovery

So if follow one of those conventions to create the MQTT topics would ease the use the in automation systems. If you like, I create an issue/feature request to file it?

zuckschwerdt:
There is already an adapter for Home Assistant in examples/rtl_433_mqtt_hass.py (converting random sensor data for auto detection needs some regex magic -- especially since a startup we do not known what sensors a user has).
Perhaps someone can use that and build a Homie auto discovery from that.

curlyel:
> adapter for Home Assistant in examples/rtl_433_mqtt_hass.py

Seems, that this is almost what I've been looking for. Quite likely a good start and/or a base for implementing a Homie adaptor...

Unfortunately, I'm not able to run it properly :-(
```
rtl_433 | ./rtl_433_dev/examples/rtl_433_mqtt_hass.py
```
connects to the MQTT broker but does not forward/convert anything.
If I got it right, it takes simple json piped to it. Running rtl_433 currently gives json output:

```
{"time" : "2019-06-17 13:05:10.307497", "brand" : "OS", "model" : "Oregon-THGR122N", "id" : 238, "channel" : 2, "battery_ok" : 0, "temperature_C" : 23.700, "humidity" : 54}
{"time" : "2019-06-17 13:05:15.376970", "model" : "sgooway2", "count" : 23}
```

zuckschwerdt:
Ah no, it's standalone. The script reads the rtl_433 ouput from the MQTT broker and then supplements the borker with discovery info.

curlyel:
> and then supplements the borker with discovery info.

Oh - understand. Let me try ...

Back to original topic:
With this flex decoder setting I'm quite reliable able to detect the PIR events:

```
decoder n=sgooway,m=OOK_PWM,s=400,l=1200,g=1500,r=12000,repeats>=5,match={25}3671538,countonly
```
This give me `sensors/rtl_433/sgooway/count 24` everytime the PIR has reported and limits the output to readings with at least 5 identical matches.

But: This will lead to a specific flex decoder setting for every of these PIR's - right?

I saw some other examples with `get=@x:{25}:id` or similar. I guess, this is something to grep out some values and treat them as "id" for example.

Do you have some hint, how to pick out the EV1527 id for using it as part of the MQTT topic if it has been repeated a couple of times. E.g.

```
sensors/rtl_433/sgooway/755765/count 24
```

curlyel:
> Ah no, it's standalone.

Have tried it now and can confirm: It is working and provide Homeassistant config topics as you said.
Unfortunately, this is not as clean and sleek as the Homie convention and brings some additional overhead to my current need... 

Though, it is a good thing, to have something like this Homeassistant adapter to ease the integration into automation systems.

For now I'm going to stick simply with your 

> `-F mqtt` ... `usechannel=replaceid`

This works perfect for me for the _protocol based decodings_ of my Oregon/Alecto/GT sensors.
I'm still looking to have some more elegant solution to pick the EV1527 id's as part of the MQTT topic or as MQTT data to overcome the need to have an individual flex decoder line for each of the EV1527 devices (for each code) in my conf, as I wrote above...

But what's available yet (means: what you have provided!) is already quite amazing.
I like the native MQTT feature! Makes it a lot easier.

Many thanks!

zuckschwerdt:
You should be able to generate an `id` with the `get` feature:  `get=@0:{25}:id:[57087288:1]` (57087288 is 0x3671538 in decimal).
Multiple ids are possible if the `match` catches them. Otherwise use multiple flex specs. Should be okay for just a few devices.

curlyel:
> Multiple ids are possible if the `match` catches them

This means: Multiple match cases in a single flex decoder line are not possible - right?
Just if a _single_ (shorter) bit pattern matches somewhere in packets from _different_ senders, than we can assign keys (channel in this case) this way:

```
decoder .... repeats>=5,rows=25,get=@0:{28}:channel:[57087287:1111 57087288:9999],match={25}0x3671538
```
_btw.: Needed to compare 28bits instead of the 25 to get the mappings..._ 

Anyway,
This generates:
```
sensors/rtl_433/sgooway2/count 23
sensors/rtl_433/sgooway2/num_rows 25
sensors/rtl_433/sgooway2/rows/0/0/len 7
sensors/rtl_433/sgooway2/rows/0/0/data 00
sensors/rtl_433/sgooway2/rows/0/0/channel 0
sensors/rtl_433/sgooway2/rows/1/9999/len 25
sensors/rtl_433/sgooway2/rows/1/9999/data 3671538
sensors/rtl_433/sgooway2/rows/1/9999/channel 9999
sensors/rtl_433/sgooway2/rows/2/9999/len 25
sensors/rtl_433/sgooway2/rows/2/9999/data 3671538
sensors/rtl_433/sgooway2/rows/2/9999/channel 9999
sensors/rtl_433/sgooway2/rows/3/9999/len 25
sensors/rtl_433/sgooway2/rows/3/9999/data 3671538
sensors/rtl_433/sgooway2/rows/3/9999/channel 9999
<...cut...>
sensors/rtl_433/sgooway2/rows/23/9999/len 25
sensors/rtl_433/sgooway2/rows/23/9999/data 3671538
sensors/rtl_433/sgooway2/rows/23/9999/channel 9999
sensors/rtl_433/sgooway2/rows/24/57087480/len 25
sensors/rtl_433/sgooway2/rows/24/57087480/data 36715f8
sensors/rtl_433/sgooway2/rows/24/57087480/channel 57087480
sensors/rtl_433/sgooway2/codes/0 {7}00
sensors/rtl_433/sgooway2/codes/1 {25}3671538
sensors/rtl_433/sgooway2/codes/2 {25}3671538
sensors/rtl_433/sgooway2/codes/3 {25}3671538
<...cut...>
sensors/rtl_433/sgooway2/codes/23 {25}3671538
sensors/rtl_433/sgooway2/codes/24 {25}36715f8
```
This currently creates a bunch of MQTT publishs  (>50), while just _one_ would be enough ;-)
as something like this for example:

```
sensors/rtl_433/sgooway2/9999/count 23
```
Unfortunately, the `countonly` does not include the `channel` anymore. 

Currently I see only the option with multiple decoder lines, one per PIR and just give them different names ("sgooway1", "sgooway2", etc.) to distinguish between them later in the home automation...

Do you have some other advise, how to shorten the output?

zuckschwerdt:
`countonly` removing the `get` mapping is likely a bug, I'll look into that.
The codes are actually 24 bit with 1 bit sync. A get of `@0:{24}` or `@1:{24}` would fit better what's "really going on".

curlyel:
> `countonly` removing the `get` mapping is likely a bug, I'll look into that.

Did you had chance already looking into it? Should I file an issue for tracking?

curlyel:
> `countonly` removing the `get` mapping is likely a bug, I'll look into that.

Hi Christian,
has this topic been addressed meanwhile?

zuckschwerdt:
Still needs to be done. The `match` and `countonly` logic need a bigger rework. Probably best to replace `countonly` with `unique`.

curlyel:
> replace `countonly` with `unique`.

I guess, this produces a single (MQTT-) output topic for repeated decodes - right?

zuckschwerdt:
Yes. Currently `match` will match every code and not account for identical codes. Not great for MQTT…

curlyel:
O.k. got it. That solves the issue for me.
Thanks again Christian!

zuckschwerdt:
Wait, i still need to add something like `unique`;)

zuckschwerdt:
The `unique` option is now supported with merbanan/rtl_433@cce4b26

curlyel:
Thank's again. Just for completeness here:
Works as expected and omits repetitions of the the same decode.

btw.: Is there an option already to issue a "dummy" token to the output. E.g.

```
get=:command:[:ON],

or

get=:command:[:1],
```

This would be useful for devices like PIR's which do nothing but sending an "ON" command when detecting movement.

Sure, we can match on 'count' being greater than 'x' to interpret this as a 'ON' command at the backend (so do I at the moment). 

But if we publish direct an 'ON' or '1' on a dummy topic no further logic is needed at the (MQTT-) client.

zuckschwerdt:
Good idea. The getters need some restructuring anyway. In the mean time try
```
get=@0:{1}:command:[0:ON 1:ON]
```
which just takes the first bit and always outputs `ON`.

curlyel:
Ah - yes ;-)
... and works:
```
sensors/rtl_433/P255/C10/time 2019-08-29 15:49:47
sensors/rtl_433/P255/C10/count 25
sensors/rtl_433/P255/C10/num_rows 25
sensors/rtl_433/P255/C10/len 25
sensors/rtl_433/P255/C10/data 124578
sensors/rtl_433/P255/C10/channel 10
sensors/rtl_433/P255/C10/command ON

```

Cool!

curlyel:
> In the mean time try

Adding this to my flex decoder setting for the EV1527-PIR made it super-easy to attach it to openHAB:

` get=@0:{1}:command:[0:OPEN 1:OPEN],`

_openHAB channel definition:_

` Type contact: motion1 "Bewegung1"     [ stateTopic="sensors/rtl433/P255/C10/command" ]`

_openHAB item definition:_

`Contact Motion1 "Bewegung"  { channel="mqtt:topic:MY-MQTT:RTL_433:motion1", expire="10s, state=CLOSED" }`

_openHAB log:_

```
2019-08-29 18:51:52.328 [vent.ItemStateChangedEvent] - Motion1 changed from CLOSED to OPEN
2019-08-29 18:52:02.726 [vent.ItemStateChangedEvent] - Motion1 changed from OPEN to CLOSED
```
When the PIR detects movement, `OPEN` is sent and after the configured `expire="10s..."` the item is set back to closed. Even works with re-triggering within the `expire`-timeout.

No hassle with topic transformation etc....
Quite happy with that!

zuckschwerdt:
Short and elegant. Perhaps you can add that as a OpenHAB howto to our wiki?

curlyel:
https://github.com/merbanan/rtl_433/wiki/How-to-integrate-rtl_433-sensors-into-openHAB-via-MQTT
;-)

curlyel:
FYI: I've made some promotion on the openHAB community forum too:
https://community.openhab.org/t/tutorial-rtl-433-brings-433mhz-sensors-to-openhab/81018

PR #290, opened 2019-06-29
Add LiftMaster 4330E recordings

ondrej-zary:
I've recorded some data from LiftMaster 4330E - a simple 1-button fixed-code garage door remote control. There are also 2-button and 3-button versions but I don't have any of them.


zuckschwerdt:
The Pulseview shows 3 different OOK pulse length, 11 "pulses" at a glance. But the codes vary in a single burst -- looks like alternating payloads. Possibly some coding with 3 symbols, but havn't seen this so far. You should transcribe the pulses from e.g. `-w OOK:-` and analyze that (perhaps quantize by 800 µs -- length then would be 1, 2, and 3).

zuckschwerdt:
PCM will also read this (`-X 'n=name,m=OOK_PCM,s=800,l=800,r=3200'`), but I wouldn't say it's meant to be PCM -- looks like always 11 pulses but variable "PCM" bit count.

ondrej-zary:
The remote should have an unique fixed code (receiver has a "learn" button according to documentation).

zuckschwerdt:
Alternating two codes could also work as fixed code, with the other one for compatibility or such.
The [codes look like this](http://triq.net/bitbench?c=%7B48%7Db85c66661130&c=%7B51%7Deee1309985c20&f=v%0A) (repeated a number of times).

PR #297, opened 2019-08-11
Test capture for Schrader 13348393 TPMS

smnrgrs:
Schrader TPMS sensor used for a variety of mainly GM cars. I haven't managed to find any information about the protocol as there's no FCC number.


zuckschwerdt:
Yes, it is similar to the two Schrader decoders we already have. OOK MC at 120 µs, use `rtl_433 -s 1024k -X 'n=TPMS,m=OOK_MC_ZEROBIT,s=120,r=300'` to read the codes. I'll have a look if the data is also similar.

zuckschwerdt:
Oh wait, it does work with the Schrader Electronics EG53MA4 decoder and the data looks sensible, right?

smnrgrs:
Thanks for your help. I don't think it works with the existing EG53MA4 decoder, but I'm a newbie so may be trying incorrectly.

zuckschwerdt:
Just running `rtl_433 -s 1024k yoursample.cu8` on the captured files works. Live capture with the default sample rate should also work.
I have renamed the sample directory to `Schrader_TPMS_PA66GF35` as 13348393 seems to be the Opel brand OEM number.

smnrgrs:
Awesome, thanks.

PR #298, opened 2019-08-20
WA216 samples

pquan:
These are samples from a WA216 wireless alarm sensors: PIR and DOOR.
they transmit in the 868 frequency band but cannot be decoded. The transmission seems very simple, yet, I failed demodulating it.

Help needed. Thank you.


zuckschwerdt:
Looks like you hit the lower frequency dead center. (View in http://triq.net/iqs) A center frequency of e.g. 868.38M would be better. But there seem to be other problems with our FSK decoding still.

pquan:
Yes, I hit the lower frequency on purpose as that was the only way to get some signal recommendation from -A. It was bogus, however.

I'm attaching a new capture with your suggested freq. I get no decoding, even by precisely timing the demodulation with s=850,l=850,r=3000.

I'm new to this project, so I'd like to ask how do we proceed from here? Anythink I can do to help with the demodulation issue ?
[samle_868_38.zip](https://github.com/merbanan/rtl_433_tests/files/3524503/samle_868_38.zip)

zuckschwerdt:
You can use Pulseview (from Sigrok project) with the `examples/sigrok-open.sh` script to get some insight. We need to find out why the decoder doesn't pick up on this signal. Will take some time.

pquan:
Yes, in pulseview the signal looks nice. I tried reading the decoder part, but altough being fluent in C++, I lack the necessary background in dsp. Sorry not being able to help. Hope you find the issue as more and more sensors are not using 433 band and/or AM modulation.

PR #299, opened 2019-08-21
Lacrosse Tx14r

spinnerdog:
added signals received while Antenna was removed and receiver was about 10cm from Lacrosse Tx14r


zuckschwerdt:
This will get the raw data from the signal
```
rtl_433 -X 'n=TX14R,m=OOK_PWM,short=264,long=508,sync=752,reset=2000,gap=550,invert'
```
It's 65 bits plus 1 trailer bit (seen as data bit). It's not clear where exactly the data fields are. Would be best if you can record some single bucket tips, so we can observe the lowest order bit changing. Also more than 256 tips would be great, so we can identify where the high byte is.

How many tips per mm water? Perhaps you can slowly pour a known amount in.

And we can possibly spot the ID field when you reset the battery (remove for a few seconds).

(Just the code lines `{66} ...` with comments are enough to develop this).

PR #304, opened 2019-09-18
Add Honeywell 5816 samples

adamweiss:
These are samples from a Honeywell 5816 door sensor. These failed to be decoded by rtl_433.


zuckschwerdt:
Thanks!
The first two samples can be read with
`-X 'n=Honeywell_5816,m=OOK_PWM,s=172,l=312,r=250'`
But the preamble is mangled and the last two sample are not read be the demod at all.

PR #306, opened 2019-09-19
Support request for SmartHome remote control

esp32wrangler:
I would like to request support for this remote control.


zuckschwerdt:
`rtl_433 -A` tells us that it's PWM with a sync pulse. `sigrok-open.sh` confirms that. The suggested flex decoder works:
```
rtl_433 -X 'n=SmartHomeDC305A,m=OOK_PWM,s=364,l=724,y=4800,g=900,r=11000'
```

esp32wrangler:
Thank you! (When I tried the flex decoder, it didn't work for some reason, but now it does.)
I really appreciate the quick response!

zuckschwerdt:
If you figure the data fields out (is there a sync word? where is the id, what bits are the buttons?) you can extend the flex spec. See the [conf/](https://github.com/merbanan/rtl_433/tree/master/conf) folder for examples. PR a documented example of your conf file if you like!

PR #312, opened 2019-11-10
Lunvon Remote Controller 

spnk0:
I made samples of a lunvon remote controller hoping to get it supported by rtl_433.


zuckschwerdt:
Great, thanks! We need to watch the size of this repo, can you remove the device_n screenshots and crop the captures to the say 3 good files (preferably the smaller ones if those show enough information, check with https://triq.org/iqs/ and maybe our pulseview script).

zuckschwerdt:
To read this device use the `-A` mode to get an idea about timings, then write a flex spec like the examples here: https://github.com/merbanan/rtl_433/tree/master/conf

PR #319, opened 2019-12-08
Add Auriol AFW2A1 test files

LiberationFrequency:
Lidl Auriol AFW 2 A1 sensor.
IAN 311588


zuckschwerdt:
Thanks! We need to watch the size of this repo ;)
Can you
- reduce the sample files to ~5 interesting samples
- crop, scale and compress pictures (best case 50k-100k each)

Perhaps you can host the high-res pictures and more samples somewhere and put a link in the README.

LiberationFrequency:
Okay, I'll do that. I'll let you know.

LiberationFrequency:
It's done. But I only changed the images from the sensor, the others I swapped out to my own repo.

PR #321, opened 2019-12-29
Signal capture for CAME TOP432S Remote Control

rbirkby:


zuckschwerdt:
Thanks! We need to watch the size of this repo ;) Can you copy the most interesting bits of info from the pdf as text to the readme and replace it with a link?
Simple button remotes are best implemented as flex decoder conf, have a look at the conf folder for examples.

PR #323, opened 2020-01-06
Elster Power Meters

gcormier:
First few captures, will run overnight.

For https://github.com/merbanan/rtl_433/issues/1196


zuckschwerdt:
Do we have good tests files to merge?

PR #325, opened 2020-01-18
lacrosse_tx145wsdth

weberzach:
I've added my cu8 files from a lacrosse tx145wsdth.  Let me know if I need to do anything else.  Once I powered the unit, many cu8 files were made quickly.  I does appear there are other lacrosse sensors in the neighborhood (Local sams club sells recently had a big display a sale on these).


zuckschwerdt:
Thanks. It'll be some days until I get around to analyze this. Open an issue for this sensor at rtl_433 and link this if you like.

zuckschwerdt:
I've split the data from 3 different transmission types and removed a 4th (weak) transmission.
This will be a bit hard to analyze. Can you remove the antenna from the receiver and place the sensor at 10cm to the receiver and record that? Just so we can decide which signal is the actual sensor here ;)

zuckschwerdt:
Ah well, the second set seems to be
```
model     : LaCrosse TX141TH-Bv2 sensor            Sensor ID : c7
Channel   : 00           Battery   : OK            Temperature: 4.10 C       Humidity  : 88 %          Test?     : No
```
and the thrid
```
model     : LaCrosse TX141TH-Bv2 sensor            Sensor ID : 57
Channel   : 00           Battery   : OK            Temperature: 2.50 C       Humidity  : 99 %          Test?     : No
```
Does that fit? If I read that right you expected 22 C (72 F) -- but I take it you had the sensor indoors.

weberzach:
> Ah well, the second set seems to be
> 
> ```
> model     : LaCrosse TX141TH-Bv2 sensor            Sensor ID : c7
> Channel   : 00           Battery   : OK            Temperature: 4.10 C       Humidity  : 88 %          Test?     : No
> ```
> 
> and the thrid
> 
> ```
> model     : LaCrosse TX141TH-Bv2 sensor            Sensor ID : 57
> Channel   : 00           Battery   : OK            Temperature: 2.50 C       Humidity  : 99 %          Test?     : No
> ```
> 
> Does that fit? If I read that right you expected 22 C (72 F) -- but I take it you had the sensor indoors.

Those are definitely outside, and not mine.  We've gone from 3.3 to 0 C today.  Those must be neighbors, perhaps without the wind speed/direction.

I'll remove the antenna and see about recording more.  Then I'll do another pull request.

zuckschwerdt:
No need for another PR, just drop a file you're sure is from your sensors here as zip -- I did clean up and merge this PR already.

PR #329, opened 2020-01-26
add Abarth 124 Spider recordings

TTigges:
Recordings of TPMS sensors on an Abarth 124 Spider.
The signals are interpreted as Jansite TPMS but show wrong values.
Byte 3 and 4 seem to be the ID.
Byte 5 seems to be a status.
Byte 6 in dec. x 1.4 is pressure in kPa.
Byte 7 in dec - 50 is temperature in °C.
No CRC?


merbanan:
Manually merged some samples. Closing issue.

PR #331, opened 2020-02-01
test files for insteon

evilpete:
Here are test files for testing insteon's  home automation protocol
(expect a pull request for the device in a day to so)


zuckschwerdt:
Great work! It'll take some days until I have time to sort and merge PR's.

evilpete:
updated Insteon sample files

zuckschwerdt:
insteon/04/g001_915M_1000k.cu8 reports 'Group Broadcast Message' instead of 'Group Cleanup Direct Message'. Is this correct?

evilpete:
the packet interpretation  looks good and I double checked the message_type  lookup table [line 335](https://github.com/merbanan/rtl_433/blob/master/src/devices/insteon.c#L335) and it seem correct with 3rd party sources

quick google Ref:
http://madreporite.com/insteon/plm_basics.html#:~:text=The%20breakdown

zuckschwerdt:
That means https://github.com/merbanan/rtl_433_tests/blob/master/tests/insteon/04/g001_915M_1000k.json is some left-over and I can replace it, right?

evilpete:
yea, I would guess leftover from a earlier version

PR #335, opened 2020-02-29
Add Malibu Boats Surf Band tests

bored-engineer:
Let me know if you need any additional samples/information or if I didn't follow the instructions correctly


merbanan:
I have question. Why would anyone ever create a product like this ?

bored-engineer:
An unauthenticated remote control that can change the speed/direction of a 10,000 pound boat? What could possible go wrong?

PR #337, opened 2020-03-21
Add samples for Honeywell CM921/BDR91/Evohome.

wintersteiger:
To go with https://github.com/merbanan/rtl_433/pull/1336.


zuckschwerdt:
Can you select just a few interesting samples? Aim for 10 or so if possible.

PR #338, opened 2020-03-22
LaCrosse TX29 - bad CRC

RoEdAl:
Frames with temperature about zero (-0.3 to 0.3) are usually reported as having wrong CRC. This is example of such frame - sequence of 9 consecutive zeros is decoded as 8 zeros.

The solution is really simple - just change short_width and long_witdh to more accurate value - 58us.


zuckschwerdt:
Thanks. The PCM is able to auto-sync but needs a decent preamble for that. Maybe I'll add more code to also allow a very short preamble if it matches close enough.

rey42:
I can confirm: bit duration for IT+ devices is 58µs, not 55.

zuckschwerdt:
Even with a short preamble the exact timing is now recognized by the pulse slicer and appropriately updated:
```
rtl_433 -R 76:vv gfile-tx29-about-zero_1000k.cu8
...
LaCrosse TX29IT, TFA Dostmann 30.3159.IT Temperature sensor:
 Exact bit width (in us) is 58.11 vs 55.00, 38 bit measured
```

PR #341, opened 2020-04-11
SMC5326

x0urc3:
Add test file and description for SMC5326


x0urc3:
Config option file has been added in merbanan/rtl_433#1346

PR #342, opened 2020-04-13
New device recording for Visonic Powercode device

maxlock:


zuckschwerdt:
Thanks! Can you rename the samples to include sample rate, frequency, and file type and change the plus to a dash (e.g. `g001_tamper-close_433.92M_250k.cu8`).

maxlock:
Files renamed :)

zuckschwerdt:
Just to be sure you really did use the defaults (no -f, no -s), right?

maxlock:
Yes, thats right, I ran rtl_433 with just a -w option to write out to a file.

zuckschwerdt:
Ah, ok. That explains why the files are that big. We usually want files from the -S option to just get one frame (message) per file. Can you possibly split the files? We need to watch the size of this repo, so just 2-4 interesting messages would be best.

maxlock:
Ahh, ok I've re-recorded the signals with the -S option, they're a bit smaller now :)

Mindavi:
I've verified that the decoder merged in https://github.com/merbanan/rtl_433/pull/1349 works for these sample files.

PR #352, opened 2020-07-20
Auriol

evilpete:
moved New directory Auriol/AFT77A into preexisting directory auriol


zuckschwerdt:
Case-changes won't reliably work. I'll pop the last merge and change it in-place.

PR #354, opened 2020-07-25
Acurite Atlas Weather Station Sample Messages

rct:


rct:
@zuckschwerdt - Are you still merging samples into the tests repo?

zuckschwerdt:
Yes, slowly. Every merge needs manual intervention. We really need a better system.

zuckschwerdt:
E.g. this: there are control chars in the README and the path should really be a 2-digit number, not "Acurite_Atlas_no_lightning_detector", with a README explaining the samples.
But I'll merge now and clean up later…

rct:
Ok thanks. I didn't see a good precedent to follow for two different "variants" of the same device and noted the old numbered directories for the different devices went away.

zuckschwerdt:
I cleaned up some stuff, but it's not perfect yet. The levels are: an optional group (e.g. manufacturer), then the devices, then numbered recording sets (with a README at each level). You can see how the result should come out in https://triq.org/explorer/
Try "smart" as quicksearch and look at the XAS_4010_Smart for an example where this works: "Intelbras" as group (headline), four sets each with a good description.

rct:
The saved signal explorer looks very cool!   I see what you are struggling with to get the tests repo in shape to be automatically processed.

zuckschwerdt:
Thanks! I still need to get around to build a similar view of the decoders, with drill down to docs, code, samples and then add the option for anyone to comment and e.g. upload more samples right there. Is going to be slow, small steps ;)

PR #358, opened 2020-08-14
Update refresh script, refresh some files that required refreshing

Mindavi:
This patch updates the refresh script to accept custom binaries. May be handy for people that don't have rtl_433 installed but want to do a quick check (I don't have it installed on my dev machine either).

A bug in the refresh script was fixed too, it did previously not accept the 16:no-wh5 'quirk' (or whatever protocol it was). It does now.

I've added some `protocol` files to prevent false positives. I'm not sure if this is the way to go? Do you have an opinion on this @zuckschwerdt? I was thinking that it would also be nice if we put the `model` name inside the folder to check, instead of the `protocol` file which contains a number that doesn't say much.

Lastly, I found some files that generated more or less lines.

- Some had duplicate lines with a same timestamp and the same data
- Some had more lines with different data (which is weird to me, as the model was the same)
- Some had less lines where it seemed like some parts of the file were not decoded anymore (deserves a closer look)


Mindavi:
Whoops, I should fix tests/prologue/02/prologue_191_60_ch2.json before this is merged, but please have a look anyway :).

zuckschwerdt:
I don't even remember adding this script. I guess it was intended for maintainers, as we don't want to update the json outputs but rather only use them for regression testing?
We should use the protocol files as last resort if a decoder is disabled. If we use them to limit the tested decoders we won't spot any false positives.
More metadata is always nice, I guess users will rather want a more integrated showcase than digging in this repo though.
I'm working on enhancing our explorer gradually as time allows. Try https://triq.org/explorer/ drill down to some set and click a (green) sample there.

Mindavi:
I'm not fully sure what you mean? I think we want to update the json files whenever a decoder adds or rearranges fields.

As far as I understand this repo is meant to help finding regressions in decoders (e.g. it stops working, it decodes fields incorrectly). It is not directly meant to find false positives, although we could add such functionality (I think an extra script would be most logical for that sort of functionality).

I think limiting decoding to the protocol it's meant to decode helps reducing noise in the json files, as we don't really care about other decoders when looking for regressions in a particular decoder.

I'm not using this repo as a showcase, rather to check if rtl_433 does not misbehave after someone changes a decoder. I also try to check if the decoder someone has written actually decodes the signal submitted to here. I also used this to generate bitbuffer files for fuzzing.

What do you mean with 'more metadata is always nice'?

zuckschwerdt:
Sorry for the quite random and unclear comment.

I use e.g. `for f in tests/acurite/Acurite_590TX/*.cu8 ; do rtl_433 -c 0 -F json -r $f >${f%.cu8}.json ; done` and then carefully select the changes I want with git. If you prefer the script that's fine. We just shouldn't apply blanket updates, but I guess you didn't want that anyway :)

To me the outputs need not reflect what some current rtl_433 actually decodes (that kind of documentation can go in a tool like the explorer) but the subset we want to test against. On that basis I don't like to add a `protocol` file if not really needed.

I consider `model` as metadata in the samples, as opposed to the `protocol` number which is just some override (like `samplerate`).

TL;DR any update and fix appreciated, I just don't think protocol overrides are needed/wanted.

Mindavi:
Let's leave it at this for now then :).

Mindavi:
Any more comments or changes needed for this to be merged?

Mindavi:
(I don't have push access to this repository, only to rtl_433)

merbanan:
@Mindavi invited

PR #366, opened 2020-10-05
iSmartAlarm Sensor Data

marc25v:
Adding iSmartAlarm door sensor data.


zuckschwerdt:
All files and descriptions look good. But strangely (as you noticed) rtl_433 doesn't pick up the modulation here. It's clearly visible in the spectrogram, looks like simple FSK.
Maybe the center frequency is too close? Not a solid bet, but perhaps try with `-f 907.6M`

marc25v:
Tried with 907.6, still giving me even if I can clearly see the signal in TRIQ:

```
Detected OOK package    @0.351659s
Analyzing pulses...
Total count:    1,  width: 25.16 ms             (25163 S)
Pulse width distribution:
 [ 0] count:    1,  width: 25163 us [25163;25163]       (25163 S)
Gap width distribution:
Pulse period distribution:
Level estimates [high, low]:  15986,     38
RSSI: -0.1 dB SNR: 26.2 dB Noise: -26.3 dB
Frequency offsets [F1, F2]:   21120,      0     (+322.3 kHz, +0.0 kHz)
Guessing modulation: Single pulse detected. Probably Frequency Shift Keying or just noise...
```

marc25v:
Also tried at 908.1

zuckschwerdt:
Strange. Could it be the harmonics (visible as "shadows") that interfere? Maybe @merbanan wants to take a look.

marc25v:
I'll drop some additional samples at different frequencies if that helps.
[iSmartAlarm Addtl Sample.zip](https://github.com/merbanan/rtl_433_tests/files/5329253/iSmartAlarm.Addtl.Sample.zip)

zuckschwerdt:
The signal really is FSK but with a very small deviation (narrow bandwidth) of maybe 10-15k. This is beyond the capabilities of our FSK demod. Perhaps by centering exactly on the signal and using a very low bandwidth something can be read, but this currently isn't going to work with rtl_433. Sorry.

PR #368, opened 2020-10-17
Sample for YG2 Smoke Sensor

MichalBartos:


zuckschwerdt:
Looks similar to the commonly used EV1527. Use `-X 'n=Emastiff-YG2,m=OOK_PWM,s=300,l=1000,r=1200'` to show the data and `-X 'n=Emastiff-YG2,m=OOK_PWM,s=300,l=1000,r=1200,match=456316,countonly'` to get this sensor only.
PR a .conf file like the ones we already have with these parameters and some description if you like.

MichalBartos:
Thanks 👍

Issue #359, opened 2020-08-15
RFC: Remove false-positives and don't test for them


Mindavi:
I'd like to adapt the `run_tests` and `refresh_all` scripts to leave out any false positives. This would ensure the `.json` reference files don't contain any output from decoders that are not the one being tested.

Besides that, a script could be created that points out false positives and gives a nice overview of which decoders give most false positives. This would prevent false positives from being invisible, without polluting the reference files per se.

Please let me know what your thoughts are about this. I'm not fully sure on the details, but I think there's room for improvement and I'd like to document what improvements are desired and/or needed.

Mindavi:
I've made an initial branch for this: https://github.com/Mindavi/rtl_433_tests/tree/feature/ignore-false-positives

PR #370, opened 2020-10-31
Show an overview of false positives in the bin/run_tests.py script

Mindavi:
The overview looks like this:

```
WARNING: LaCrosse-TX35DTHIT generated 1 false positive(s) in other decoders: {'LaCrosse-TX29IT'}
WARNING: Acurite-609TXC generated 6 false positive(s) in other decoders: {'Oregon-SL109H'}
WARNING: Oregon-SL109H generated 1 false positive(s) in other decoders: {'Prologue-TH'}
WARNING: Nexa-Security generated 157 false positive(s) in other decoders: {'KlikAanKlikUit-Switch', 'Proove-Security'}
WARNING: Proove-Security generated 80 false positive(s) in other decoders: {'KlikAanKlikUit-Switch'}
WARNING: DSC-Security generated 6 false positive(s) in other decoders: {'Schrader'}
WARNING: Fineoffset-WH0530 generated 1 false positive(s) in other decoders: {'GT-WT02'}
WARNING: Akhan-100F14 generated 193 false positive(s) in other decoders: {'Waveman-Switch'}
WARNING: Abarth 124 Spider generated 1 false positive(s) in other decoders: {'Renault'}
```

Related to #359. This only shows an overview, I have an update for refresh_all as well that removes false positives from the generated JSON files.


bin/run_test.py
```diff
@@ -119,7 +101,21 @@ def main():
...
                 continue
             try:
-                results.append(json.loads(json_line))
+                data = json.loads(json_line)
+                if "model" in data:
+                    expected_model = expected_data[0]["model"]
+                    actual_model = data["model"]
+                    if actual_model != expected_model:
+                        if actual_model not in false_positives:
+                            false_positives[actual_model] = dict()
```
Mindavi, line 110: I think this could be done cleaner, but not sure how.

zuckschwerdt:
I'm not a fan of splitting out those 10 lines into a "module", I'd rather have the scripts self-contained and that small helper easy to find in the same source.

Otherwise more warnings and summaries is always great!

Mindavi:
Let's go for a version without that split then :).

PR #372, opened 2020-11-08
Improve refresh_all script to ignore false positives and remove false positives from expected json files

Mindavi:


tests/lacrosse/tx29-it_tx35dth-it/gfile002-TX35DTH-IT.json
```diff
@@ -1,2 +1 @@
-{"time" : "@0.127256s", "brand" : "LaCrosse", "model" : "LaCrosse-TX29IT", "id" : 10, "battery_ok" : 1, "newbattery" : 1, "temperature_C" : 23.800, "mic" : "CRC"}
```
Mindavi, line 1: It seems like both signals are in this file, so maybe it would be good to split it up instead of removing this expected output.

zuckschwerdt: Always good to have "bad" files as tests to make sure corner cases work. But yeah, this output won't fit into any test logic, likely best to split the cu8.

Mindavi: I've just removed the updated file for now, but it would be good to handle this better in the future.

PR #373, opened 2020-11-12
Feature/remove old ignore files

Mindavi:
Assuming these ignore file were added when the DSP was not good enough or when the protocol was not yet merged.

With the ignore file removed, all these tests pass on my side, so I think it's good to remove them.

There are also some ignore files that can probably be removed, but the decoders are now generating different output from when the ignore files were added (mostly extra fields). Those deserve a closer look.

Is this the correct approach? I'd think that all is ok when the decoders work properly, but maybe I don't understand the purpose of the ignore files fully.


merbanan:
@zuckschwerdt looks good to me.

zuckschwerdt:
Feel free to clean things up. I mostly add ignore files when the samples PR comes with json files but the decoder isn't merged.

PR #375, opened 2020-11-22
Add Salus RT300RF Thermostat Samples

mcginnly:
Some samples here for the RT300RF thermostat. 
I've included the manual and the datasheet for the RF transmitter.
This is a useful link for someone who has hacked a slightly later model, although the signal looks very similar in its construction and both thermostats use the same transmitter. https://the.cyclingengineer.co.uk/2013/11/23/home-automation-integrating-salus-rt500-rf-in-openhab-using-a-jeelink/ 
Whilst I've managed to get some appropriate data out of the samples using 
`rtl_433 -R 0 -X "name=Salus300RF, m=FSK_PCM, s=416.6666, l=416.6666, r=33000" -r <filename>`
it doesn't seem to work on the live feed with 
`rtl_433 -R 0 -X "name=Salus300RF, m=FSK_PCM, s=416.6666, l=416.6666, r=33000" -f 868.286Mhz`

I'd be enormously grateful to know why and also be pointed to how to construct a decoder.
I've added a conf to `/usr/local/etc/rtl_433/SalusRT300RF.conf`
And added a device to `/usr/local/include/rtl_433_devices.h` but it doesn't seem to have worked, but then I'm kind of blind without a 'how to'.

The conf is:
```
r_device SalusRT300RF = {
        .name           = "Salus RT300RF",
        .modulation     = FSK_PULSE_PCM,
        .short_limit    = 416.6,
        .long_limit     = 416.6,
        .reset_limit    = 33000,   //no idea what this should be?
        .json_callback  = &salusRT300RF_callback,
        .disabled       = 0,
        .demod_arg      = 1,
        .fields         = output_fields
};
```


zuckschwerdt:
The difference between the tow commands (file vs. live) is the demod. Try `-Y classic` on the file and `-Y minmax` on the live.
Short/long need to be the same and can be anything around 416, there is a preamble (`aaaa`) here and slicer will lock onto the preamble in a wide range.
The longest gap here is ~7500µs (18 0's) so reset should be about 9000 maybe.

mcginnly:
Thanks Zuckschwerdt. 
Sorry for the delay, I've been trying to get the live feed to work, and work out a decode, which I've now done.

I found I got more consistency of output using s=833 and l=833 with a 16000 reset (a little less than the gap between the repeated signal bursts) which gives a straight 48 output on the live feed.
{48}2dd40001015a is the thermostat calling for heat.
{48}2dd40002025a is the thermostat off.

I decoded to a straight on/off variable and assumed the first 16 bits after the preamble is a sync word.

Here's the flex decode.
`rtl_433 -Y classic -R 0 -X "name=SalusRT300RF, m=FSK_PCM, s=833, l=833, r=16000, preamble={24}0xaaaaaa,get=@0:{16}:Thermostat ID, get=@28:{4}:heat:[1:ON 2:OFF]" -f 868.286Mhz -F "mqtt://<mqtt server address>,retain=1,devices=sensors/rtl_433/P[protocol]/C[channel]"`

How do I get this to work as a default decoder? Is it as simple as putting a conf file in my rtl_433/conf/ directory, or do I have to recompile? 

Do I need to do anything else with the tests?

Cheers

zuckschwerdt:
Yes, `2dd4` is a common sync word. Using `preamble=aaaa2dd4` will be more robust. A conf file is ideal here. Then load it with `-c` or from a main conf with include. Make a PR with the documented conf if you have time!

mcginnly:
will do. Thanks

PR #378, opened 2020-12-12
Add Acurite 515TX samples

osenft:


Mindavi:
Looks good. I'll leave this open until I (or someone else) get around generating the json files as well, otherwise they might get forgotten.

osenft:
> Looks good. I'll leave this open until I (or someone else) get around generating the json files as well, otherwise they might get forgotten.

I just added them. I had them ready, but wasn't sure if they should be in the same PR. I just decided that this is what would make most sense. PTAL.

Mindavi:
This works with the latest version in master (8cdb56d9ddfbf3cbd51bdf299e1cfbba3efd6c4f), so that's good. Thanks for improving the samples repository :)

PR #380, opened 2020-12-18
Add Blueline PowerCost Monitor samples

jbrzozoski:


merbanan:
Looks ok to me. @Mindavi any comments?

Mindavi:
Looks good to me. The only thing we need to make sure these tests actually work is a `protocol` file with the contents `176:45364`. Otherwise they fail with 'decoder gives no output'.

Let's leave this PR open until https://github.com/merbanan/rtl_433/pull/1580 is merged, then we don't forget to remove the 'ignore' file :).

PR #386, opened 2021-02-28
Acurite 5-n-1 unsupported (VN1TXCA2)

rbflurry:
I am sorry if I did this the wrong way. I have an acurite 5-n-1 (VN1TXCA2) that is being detected incorrectly as an acurite-tower. I attached a few samples.

Weather conditions as follows:

Outside temperature: 40 Fahrenheit 
Outside humidity: 99%
windspeed: 2mph
wind direction: east

FCCID: https://fccid.io/RNEVN1TXCA2


rbflurry:
After looking more into the captures I dont think I am pulling enough data.

rbflurry:
Disregard. This was a signal issue on my end.

PR #392, opened 2021-04-21
Add Tesla charge port opener sample

nunojpg:
Hello, I would appreciate if someone experienced could confirm OOK_MC_ZEROBIT is the correct decoding (while coding is not so relevant since its a fixed value regardless of how we interpret it), and if there should be any different handle for the preamble since I am confused with the row '0000'.


zuckschwerdt:
The .cu8 spectrogram (use https://triq.org/pdv/ ) looks like clean OOK.
Running the sample with `-A` [gives the pulse data](https://triq.org/pdv/#AAB0170401018C031C04A427148080808080808080808080808255+AAB02D0402018C031C04A42714808091809191809191919191809081908180908081808090819081918080809081809255+AAB02D0401018C031C04A42714808091809191809191919191809081908180908081808090819081918080809081808355). There is a preamble and then three repeats, each preceded by a de-sync of invalid 3 half-bit 0's, in each packet. Yes that's MC. But you could also read it as PCM to get the preamble and de-syncs.
A conf file flex decoder will work, there are some examples of similar "single code" protocols in the conf folder.

nunojpg:
I included already a flex decoder example in the README:

`-X "n=Tesla charge port opener,m=OOK_MC_ZEROBIT,s=400,r=1200,t=20"`

But I am not sure if there is a way to match multiple rows of data.

zuckschwerdt:
That timing spec is already optimal.

nunojpg:
```
time      : @0.203604s
model     : Tesla charge port opener               count     : 5             num_rows  : 5             rows      :
len       : 13           data      : 0000,
len       : 52           data      : 094aa9b38da19,
len       : 52           data      : 094aa9b38da19,
len       : 51           data      : 094aa9b38da18,
len       : 1            data      : 0
codes     : {13}0000, {52}094aa9b38da19, {52}094aa9b38da19, {51}094aa9b38da18, {1}0
```
Considering there are 3 rows of same data (except for the last bit), is there a way for a conf decoder to match multiple rows of data or this needs a custom function?

zuckschwerdt:
You likely want (from `rtl_433 -X`)
```
	repeats=<n> : only match if some row is repeated <n> times
		use opt>=n to match at least <n> and opt<=n to match at most <n>
	match=<bits> : only match if the <bits> are found
	countonly : suppress detailed row output
```
i.e. `,repeats>=3,match=094aa9b38da1,countonly`

nunojpg:
There is always 2 rows = 094aa9b38da19 and 1 row = 094aa9b38da18.

repeats>=3 can never work.

With flex I can make sure two match 2 identical rows and the content 094aa9b38da19. I will do it, It's good enough.

Thanks for your hints.

nunojpg:
@zuckschwerdt how  would it be do add a decoder? If add the file to rtl_433/conf/tesla_charge-port-opener.conf and this must always be selected manually? Can this .conf decoders also be registered by default?

zuckschwerdt:
There is a list of conf files read at startup. Copy the decoder there or add an include with `config_file` to one of the files.
```
Trying conf file at ...
```

nunojpg:
Decoder: https://github.com/merbanan/rtl_433/pull/1704

PR #397, opened 2021-07-10
check for false positives from simple data

evilpete:
Here is the test code noted in pr merbanan/rtl_433#1444 and issue merbanan/rtl_433#1441


zuckschwerdt:
This looks great!

zuckschwerdt:
Good to merge?

evilpete:
yup, has been...

had just added an extra option today. and fixed spelling, works great for me

***@***.***:~/Projects/rtl_433/rtl_433_tests/test_false_positives# make
./run_test.sh
/bin/bash: warning: setlocale: LC_ALL: cannot change locale (UTF-8)
No False Positives with pattern 0x00
No False Positives with pattern 0xFF
Potential False Positives with pattern 0x55
     Protocol Model
3 Prologue-TH
21 Calibeur-RF104 38 Generic-Temperature
68 Kerui-Security
84 Thermopro-TP11
86 Smoke-GS558
87 Generic-Motion
108 Hyundai-WS
114 Maverick-ET73 115 Honeywell-ActivLink
116 Honeywell-ActivLink
163 Acurite-590TX
see file 0x55_out.json for details

evilpete:
@merbanan   you might want to rerun test_false_positives,  was tracking something else down (broken La Crosse sensor) and saw a  false positives popping up.  So I re-ran the test_false_positives code and it seems a bunch have snuck back into the code. :-)

zuckschwerdt:
Thanks! The report output is clear on what to fix, again very useful!

evilpete:
> Thanks! The report output is clear on what to fix, again very useful!

@zuckschwerdt 
Before, when I fixed the first set of "false positives", I was mostly sanity tested extracted values,  a better (earlier) approach would probably just to look for bit buffers. that are **_all_**  ones or **_all_**  zeros, perhaps right after the modulation.

zuckschwerdt:
I thought so too, maybe we should not pass on boring bitbuffers, all 1's or all 0's.

PR #401, opened 2021-09-24
Add samples for the Clipsal CMR113 cent-a-meter power meter

mikey:
To complement adding support to rtl_433 for these devices, here are some samples. 

rtl_433 PR is https://github.com/merbanan/rtl_433/pull/1814


mikey:
from https://github.com/merbanan/rtl_433/pull/1814#issuecomment-926392344
> Can you add back the freq/rate info to the filenames, it's the only metadata we have.
Yep.

mikey:
Updating code to reference the correct name of CMR113. The branch name here is the only wrong thing now (hopefully).

PR #403, opened 2021-10-22
Add tests for megacode decoder

aaronsp777:
Corresponding decoder is added in https://github.com/merbanan/rtl_433/pull/1834


aaronsp777:
Is this ok to be merged in or should I make some improvements?

zuckschwerdt:
With the tight low-pass and squelch the files don't test the pulse detector. Conversion to .ook won't loose information and will be much smaller.
Can you convert all files to .ook (`rtl_433 -w thefile.ook thefile.cu8`)? I'll then add ook as input to the test script.

The scripts may be useful, but this repo isn't a good place to put them, maybe add a link to some other place or repo where you can make those available?

aaronsp777:
Just wanted to circle back.  Were you able to add ook as an input to the test scripts?

zuckschwerdt:
Support for alternative .ook input for checking .json added with 143592d now.

zuckschwerdt:
Looks good. Two notes: the `cuvoodoo` folder should be named `01`, and can you set `-f 318M` so the ";freq1" and ";centerfreq" show up correctly? (or just patch `318000000` in)

aaronsp777:
I was able to rename the directory, but despite trying various combinations of "-f 318M" I was unable to get to write the *.ook files with meaningful values of ;freq1 and ;centerfreq.  

Please note I did not record these samples nor do I have the privilege to record my own.  I am merely providing proper unit tests so that the 'megacode' decoder can continue to be tested in case of future refactoring.

All the files are generated automatically from the provide script: "make_all".

I do however think it would be in poor style for me to falsify tags in the OOK that someone else could not reproduce in the future.

Please note that the filenames already contain "318M" in the filename that match the style for of all the other tests in this repository.

If in the future rtl_433 properly sets these values when "-f 318M" is passed converting from IQ files to OOK files than these files can be regenerated automatically with by rerunning make_all.

I do however believe such work to be out of scope of adding these simple tests and it should not block the PR.

PR #406, opened 2021-11-29
Add LaCrosse TX34 rain gauge device

rey42:


zuckschwerdt:
The `.sr` should not be merged, otherwise looks good.

zuckschwerdt:
Oh, less than a kB? That's not the file we generate with e.g. `rtl_433 -w g001_868.3M_1000k.sr g001_868.3M_1000k.cu8` then, right?

rey42:
This is the file generated by rtl_433 but without the analog parts - useful for protocol example and analyzis

merbanan:
Tiny files are ok.

zuckschwerdt:
> Tiny files are ok.

Agreed. Wouldn't the .ook file be more interesting though? E.g. `rtl_433 -f 868.3M -s 1000k -w g001_868.3M_1000k.ook g001_868.3M_1000k.cu8`
See the output (with added slicer hint) here: https://triq.org/g001_868.3M_1000k.ook
Get the file and drop on https://triq.org/pdv/ to see:

screenshot

PR #408, opened 2021-12-06
tests: add DeWalt Wireless-Tool-Connect (WTC) tool-to-vaccum samples

sladen:


zuckschwerdt:
Awesome detail. You hit the carrier dead on (DC) and the signal is clipping (too loud) that's why it's harder to eyeball I/Q/AM and maybe decode.
The continuous carrier for idle/warmup is not great and could violate specs ;)

The coding is MC at ~58 µs half-bit time. Could be GFSK, that's why the timing isn't consistent. Tune away by 40k or 60k and try a bigger distance (or remove the receiver antenna).

Note that s=l for MC (NRZ PCM), it's just the (half-)bit width.

sladen:
Fresh pair of samples @ 433.90 Mhz (with less of an antenna and a bit more distance).

zuckschwerdt:
Yes, much better. Thanks! It even works with the pdv viewer (use `-A` on a file to get the link), e.g.
https://triq.org/pdv/#AAB1061BE400300044007C7DF0C2B0829292929292929292929292929292929292A2929292929292929292929292929292929292929292929292929292929292A29292929292B392929292B29392B293B2A3929292B293B3B2929392B2929292C292929292929292929292929292A2A2929292A29292929292929292929292929292A292929292929292929292929292929292A2929292B392929292B29392B293B293929292B293B3B2929392B2929292D255

Because of the carrier all data is in one big packet, we need a repeated search for `aaa9` preamble, then MC decode.
E.g. dewalt-wtc-start-clean_433.90M_250k.cu8 gives `f993ca30` twice then.

sladen:
No, no, thank *you* for the super-fast feedback.

Over the weekend being able to *see* the data in the waveform (but failing at getting any meaningful automated decode) was frustrating.  Having feedback is wonderfully satisfying!

Is there anyone of specifying the two FSK frequencies to lock on to?  (I think normally the frequencies would be close together, but in this case there is a 10x ratio between the two?)

zuckschwerdt:
You don't need to know the frequencies, they just need to be in the band (of 250k here) covered, at least for rtl_433.
The FSK deviation is about 20 kHz, which is normal, and for GFSK/MSK based on the symbol rate (broadly speaking it's so the phase aligns with the symbol changes).

SorX14:
Sorry to resurrect, but this is the only place I've found attempting to decode these signals. I can share two other WTC tool recordings (DCW200 and DCS727), but was wondering if you've made any further progress in decoding? 

Using `rtl_433 -R 0 -X 'n=DeWaltWTC,m=FSK_MC_ZEROBIT,s=55,l=135,r=1000,invert,reflect'` on my own live capture isn't repeatable - I've extracted the WTC module from one tool, leaving it running next to the RTL-SDR and the every-10-seconds decoded packets rotate through different values. 

My goal is to create a receiver to enable non-Dewalt WTC dust extractors to be used. Would a CC1101 be appropriate? I haven't had much luck so far.

PR #410, opened 2021-12-29
Added recordings of GTO Access Systems Mighty Mule FM231 Wireless Driveway Sensor

BarryMar:
I couldn't seem to get the included [MightyMule.conf file](https://github.com/merbanan/rtl_433/blob/master/conf/MightyMule-FM231.conf) in rtl_433 to work with my transmitter, so here's some recordings.

There were some errors getting generated during the recordings, so I've included some URH recordings as well if that helps.

(Please ignore the Lacrosse stuff.  I'm still learning how to deal with GitHub and need to figure out how to suppress that old stuff properly)


zuckschwerdt:
I'll have a look soon. If you search the rtl_433 issues there should be a report with exactly that bad recption problem iirc.

PR #411, opened 2022-01-19
unknown sender Eberle Instat 868r1

dottoreD:
Hello to you,
I have uploaded data and available information for my not recognized device. Some analysis have been done by myself and I tried to show data in proper way.
I hope somebody can help out as specially checksum, or what I think it's checksum, is not clear to me at the moment

best regards,
Dieter


zuckschwerdt:
Did you convert the .cu8 files? That looks really strange. There is a DC spike, strong harmonics and a high frequency noise (at band edges).
Is there a chance to get a clean signal?

zuckschwerdt:
Oh, "logging data by wire directly into an Arduino": so you do have the bitstream directly from the radio? Can you share that? The format doesn't matter, but needs to be raw (no Manchester decoding).

dottoreD:
cu8 files have not been manipulated, I have uploaded 5 more cu8 files, hopefully showing better signals. Replicates are now included. For the second signal one only 2 out of 3 replicates have been captured. Unfortunately I have no real feedback for their quality but with I/Q spectrogram I can see some signal / peaks looking not that bad.

Maybe for further explanation:
- initial tests have been done by URH and rtl_433 => identifying FSK and Manchester code
- further tests done using URH to receive digital bit stream output, both Manchester encoded and decoded => identifying concept of data stream of 6 nibbles including ID, other data and checksum
- collecting received data gave me a clue how data are organized, see file "protocol.txt"
- additional tests by wiring a test unit to Arduino to receive a good number of high quality data (without radio and SDR), actually only for "learn codes", meaning different ID but the same rest of data for all of them.
- I assumed that learn codes with minimal and regular steps might be best to reverse engineer nibble #6 with checksum, because this checksum appears to all data and other protocols have been found not giving a good dataset

The uploaded file "learn.txt" contains all that data, after Manchester decoding. Just 6 Nibbles for each sending, but it's not more than that. The file "learn_bitstream.txt" is before decoding, but strange header is cut. This was done to focus on real data set.
I saw, that the header is not exactly the same as with radio, maybe some imperfection for my code to capture data by Arduino and because I had to manually take out one protocol from 3 replications always send. I did a lot of things manual, and by Office Calc. But the real data starting always after a "000" signal, and last "0" has been taken as first bit of data.

zuckschwerdt:
In your 6-nibble-data: the first two bytes are some value, but the last byte changes with every packet. That's not a checksum but perhaps a rolling authentication code?
And with a different start offset after "000" the coding could also be DMC.

dottoreD:
This is misunderstanding: the 6-nobble-data are created by collecting data from learn mode only, and are not sorted by time sequence. Each one is a new ID created. This code is only repeated in a single learning phase and do not change at that time, none of the nibbles.
But I started re-creation of new code again and again to get out each possibility. I had expected to get 000 - FFF, but this isn't the case. Then these codes have been sorted by first 3-nibbles. Most codes have be re-created 2-3 times, and exactly the same output was created. So no random or rolling code in that phase.

Nibbles 1-3 do never change in reset or switch mode, just data starting from nibble 4 on.

For nibble#4 (B/4) I think it identifies information that this is learning code, 2 possibilities for odd/even. Next one (A/F) just parity, because the same has been found for other to identify odd/even. To be honest just my thinking but reproducible.
Nibble#6 is that what I think checksum.

>And with a different start offset after "000" the coding could also be DMC.
I don't understand what you mean. What is DMC?

zuckschwerdt:
https://en.wikipedia.org/wiki/Differential_Manchester_encoding

dottoreD:
Got it.
To be honest, I did not though a lot about different Manchestercodes. For Thomas and IEE results are just inverted when I understood right, but details for DMC I never understood. I hope, this doesn't change a lot for that case.
Here the problems just starts with the question if high or low frequency in FSK represents digital high or low.
So I just took simply a Manchester "01" as "0", and vice versa. The end of header is well highlighted with the "000", which cannot decoded, and after that 51 bits follow. So I took the final "0" to have 52 => 26bits are decoding is maximum possible number. With again some header of 2 bit that is always the same and doesn't fit into 8bit nibble scheme. My thinking.

dottoreD:
I have uploaded files for reset sequence and on/off and reset sequence for V0101

zuckschwerdt:
Yes, IEEE and Thomas are MC are just inverted. DMC is shifted by a half-bit width and will give a different decode. So if values look too random it might be DMC.

Can you open an issue at rtl_433 with a summary of findings? Discussions here with valueable info get lost :/

PR #429, opened 2022-08-03
cosmetics: Group rubicson device samples

robho:
The README proposes to group devices by the same manufacturer in
a single directory under the tests directory. There are currently two
rubicson device directories so I suggest to group these to a single
rubicson directory.

I'm about to add samples for a third rubicson device so it would be nice
to have a clear directory structure for rubicson devices.


zuckschwerdt:
Thanks!

PR #432, opened 2022-08-12
Add samples for SRSmith SRS-2C-TX

gcohen55:


gcohen55:
This is related to a rtl_433 PR: https://github.com/merbanan/rtl_433/pull/2147 that includes support for this device.

PR #440, opened 2023-01-10
Add samples for Watchman Sonic Advanced

garethpotter:
- samples
- corresponding .notes file contains expected values
- .bitbench file contains link to partially decoded data (extracted with URH)


zuckschwerdt:
Can you move all text from the bitbench and notes files into one README.txt?

zuckschwerdt:
Looks great, thanks! (yes, .md not .txt -- I was confused there)

PR #447, opened 2023-02-14
Add samples for EEZ Tire TPMS

Gliebig:
I've collected data for this device with the intent of my home assistant on my RV to be able to capture tire pressure data while it is in storage using an SDR. I'm totally new at this and just trying to be helpful to get another sensor defined. This is such a useful project.


zuckschwerdt:
Thanks! That's a great analysis and write-up. Love your perfect testing setup.

Quick notes: rtl_433 Windows releases are at https://github.com/merbanan/rtl_433/releases
The "NRZ" is actually RZ (what we usually name PCM), the same as the raw Manchester bits, see https://triq.org/rtl_433/PULSE_FORMATS.html

I'm not sure the signal is really ASK/OOK you did hit the signal dead center and so one frequency component (of FSK) might be lost. Also there is very much clipping (too loud). Drop a .cu8 sample on https://triq.org/spectrogram-next/ and scroll zoom to inspect the waveform.
Best to remove the antenna or get the receiver father away and offset the tune, e.g. `-f 433.88M`

Open an issue "Add support for EEZ Tire TPMS" at https://github.com/merbanan/rtl_433 to discuss.

PR #448, opened 2023-02-17
Add SWETUP garage door opener

FoxWhiskey:


zuckschwerdt:
Thanks for the sample files and very thorough description.

The compatibility and analysis suggests a PT2260/PT2262 or EV1527 encoder /protocol.

You can just copy and modify one of the [rtl_433 .conf files](https://github.com/merbanan/rtl_433/tree/master/conf) with similar settings to the flex (-X) parameters found.

zuckschwerdt:
Both are practically the same.

PR #451, opened 2023-03-18
Add: samples for Silvercrest Doorbell T

tnixeu:
This PR adds samples for a the Silvercrest Doorbell T (STKK 16 B1) 

Model number: STKK 16 B1
Manufactured: 2022-09
IAN: 498825_2204

I needed to set the sample rate to >1000k.


zuckschwerdt:
Interesting as this is not a x1527 encoder but energy harvesting with a faster PCM protocol.

Samples are on DC and clipping, it should be `README.md`, resample pictures to be 50k (or if there is much detail 100k).

zuckschwerdt:
Files recorded at 433.92M will not work. Please record e.g. at `-f 433.90M`

tnixeu:
I have replace the files with recorded at `-f 433.9M`. Is there a difference between `-f 433.9M` and `-f 433.90M`. I only now noticed the extra `0`.

I am a bit curious. Is there a reason for `433.90M` instead of `433.92M` for this doorbell? Most other test files seem to be recorded at `433.92M`

merbanan:
The rtl_sdr can not handle signals that perfectly hit DC. You need to be some Hz high or low off.

tnixeu:
I broke mine somehow. So I am closing this PR.

PR #463, opened 2024-02-06
Add BMW Gen5 TPMS sample files

petrjac:
Moved my question to related issue https://github.com/merbanan/rtl_433/issues/2821


zuckschwerdt:
It's basically the same. MC can't reliably discern the timing of the short transitions until the first long one.
Smaller recordings would be prefered, cu8 or cs8 and at most 1024k sample rate.

petrjac:
My SDR (Airspy R2 https://airspy.com/airspy-r2/) only support _10M_ or `2M5_ sample rate.

Should I resample using some software tool @zuckschwerdt ?

PR #464, opened 2024-03-17
Thermor DG950 weather station need help to decode.

NicolasGagne:


zuckschwerdt:
The signal is directly on 433.920M please tune away a little (`-f 433.89M`) and also record with a higher sample rate (`-s 1024k`).
The signal is clipping badly (too loud) please use more distance or remove the antenna.
Check samples on https://triq.org/spectrogram-next/ to see that most clipping is gone (the red lines in the lower waveform display).

zuckschwerdt:
Interesting signal. The actual carrier seems to be at 433.990 MHz, so -f 433.95M could be better. But the samples are clean and work.
Please open an issue at rtl_433.

NicolasGagne:
Done #2879
Thanks

PR #469, opened 2024-05-29
Add sample files for rolleaseacmedia blind motor controller.

bdnstn:
Sample files and notes on my analysis of the signals for the rolleaseacmedia blind motor controller.


zuckschwerdt:
Good analysis and description. This can be implemented as flex decoder .conf file. Open an issue at rtl_433 if you need further help or just PR a conf file with your findings an documentation.

bdnstn:
Sure, I can write a flex decoder in a conf file. I'd then make a PR at rtl_433 (not rtl_433_test)?

PR #496, opened 2026-05-31
Add Landis & Gyr GridStream reference captures (LandisGyr-GS)

swannman:
Adds the first real off-air sample data for the Landis & Gyr GridStream decoder (`src/devices/gridstream.c`, protocols 271/272/273), companion to rtl_433 PR #3546.

### Samples — `tests/LandisGyr-GS/`

| file | subtype / CI | frame |
|------|--------------|-------|
| `g001_903.2M_250k.cu8` | 0x55 / 0x30 | broadcast beacon |
| `g002_908.9M_250k.cu8` | 0xD2 / 0x52 | encrypted directed (AES-256) |

Real captures (AirSpy R2 @ 908 MHz / 10 MSPS), frequency-translated to each frame's channel and decimated to 250 kSPS (GridStream is FHSS). Not synthetic.

### No expected-output JSON (intentional)

These GFSK frames only slice with the min-max FSK detector (`-Y minmax`); the default detector does not pick them up, and `bin/run_test.py` runs rtl_433 without `-Y`. So there is no `.json` here — the harness skips sample dirs that lack one (same as `tests/Neptune-R900/01`). Decode manually with:

    rtl_433 -r tests/LandisGyr-GS/01/g001_903.2M_250k.cu8 -Y minmax -M level

### Privacy

`g001` is a broadcast beacon from the capturer's own meter (destaddress = broadcast; only that one LAN ID present). `g002` is encrypted and carries no addressing on the wire. No third-party meter identifiers are present.

Marked draft pending #3546.


merbanan:
Can you move it into a sub-directory (01) ?

PR #501, opened 2026-06-10
Derive rtl_433 -f/-s from sample filename so FSK min/max path is tested

deviantintegral:
Testing a replay of a CU8/CU16 file ignores the capture frequency unless it is passed explicitly. rtl_433 defaults to 433.92 MHz and always selects the classic FSK slicer. This means captures above 800MHz are tested against the classic slicer, which doesn't match default rtl_433 behaviour.

This updates the test scripts so samples can include a sample rate override in the filename and sets the minmax slicer for captures it runs on by default.

### DIFF
--- a/bin/refresh_all.py
+++ b/bin/refresh_all.py
@@ -9,13 +9,16 @@ import fnmatch
 import subprocess
 import json
 
-def run_rtl433(input_fn, samplerate=None, protocol=None, rtl_433_cmd="rtl_433"):
+from sample_params import rtl433_args
+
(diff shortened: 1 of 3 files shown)


zuckschwerdt:
rtl_433 does parse the frequency from the filename already. The fix from https://github.com/merbanan/rtl_433/pull/3554 properly applies that data.

deviantintegral:
Oh! Thanks, I totally missed there was parsing too ([link for the curious](https://github.com/merbanan/rtl_433/blob/master/docs/OPERATION.md#file-names)).

PR #507, opened 2026-07-12
Add test fixtures for Ecowitt WN38 black globe thermometer

rusnac:
Real 868.3 MHz FSK captures of the Ecowitt WN38 black globe thermometer, the WBGT
component of Ecowitt's sensor lineup ([product page](https://shop.ecowitt.com/products/wn38),
[manual](https://oss.ecowitt.net/uploads/20251128/WN38AN-UserManual.pdf)). Decoder
merged in merbanan/rtl_433#3601.

Updated per review: two `.cu8` samples with reference JSON generated against rtl_433
master, plus the bitbuffer codes for all five captured readings in `01/README.md`.

All five readings come from the same device (id 0x002885, matching the "ID : 2885"
sticker in `id.jpg`) and span 24.5–31.0 °C. Each was confirmed against the sensor's
own LCD display and carries a valid CRC-8 and checksum. The frame is byte-identical
to the WN34 layout but with family code 0x38 and subtype 4 (scale 10, no −40 offset);
full field breakdown is in the README.

A note on which two captures are included: the recorded bursts are amplitude limited
(clipped), and three of the five are consequently invisible to rtl_433's default pulse
detector — g001, for example, decodes only under `-Y classic`. g005 and g006 are the
two that decode on default settings, which is why the larger g005 is included rather
than a smaller capture. The codes for the other three are unaffected, since they
bypass demodulation.

Also includes front, label, and ID-sticker photos of the hardware.


merbanan:
2 cu8 samples but all the codes pretty please.

rusnac:
Updated: reduced to two `.cu8` (g005, g006) with reference JSON generated against
rtl_433 master, and moved the bitbuffer codes for all five readings into
`01/README.md`.

One thing worth flagging while I was picking the two captures: these recordings are
clipped, and three of the five don't decode with the default pulse detector at all
— g001 needs `-Y classic`. Same signal, valid CRC either way. Happy to open an issue
with a sample if that's of interest, or to re-record at lower gain if you'd prefer
cleaner fixtures.

merbanan:
Pretty please samples without clipping.

merbanan:
Or not, if the classic detector can take it we are good.

PR #508, opened 2026-07-12
Add test signals for Fine Offset / Ecowitt WH52 3-in-1 soil sensor

Verstreubulator:
Adds sample captures for the Ecowitt **WH52**, a Fine Offset soil probe that reports
soil moisture (%), soil temperature, and electrical conductivity (EC, µS/cm). It shares
the WH51's FSK/PCM modulation and `aa2dd4` preamble but uses family byte `0xA2` and a
24-byte frame, so the WH51 decoder doesn't pick it up.

Under `tests/fineoffset/wh52/01/`:

- Four `.cu8` captures (US 915 MHz, 250 kHz), one per physical unit, recorded live
  off-air next to the manufacturer gateway/app for ground truth.
- A `README.md` with the id / moisture / temperature / EC / battery for each file, and
  signal + timing notes.

The four units span moisture 25–44 % and EC 6–107 µS/cm. A companion decoder PR is
opened against merbanan/rtl_433; each capture verifies with it via `rtl_433 -r <file>`,
and the decoded values match the app.

These are shared as a starting point in the hope they help others; notes reflect what we
observed and how we checked it, not a definitive spec — corrections welcome.

Captured using the RTL-SDR "rtl_433 (next)" Home Assistant add-on by Peter (@pbkhrv,
github.com/pbkhrv/rtl_433-hass-addons) — thanks to him, and to Benjamin Larsson (@merbanan)
and the rtl_433 contributors, whose work this builds on.


merbanan:
2 cu8 samples with as different values as possible and lots of codes would be nice. Look in the git history how it looks like. And json decode files also.

Verstreubulator:
Thanks for the pointers — restructured to the repo layout: tests/fineoffset/fineoffset_wh52/ with gfileNNN.cu8 + matching gfileNNN.json decode files. There are now two multi-frame samples (each concatenates two physical units), so each file yields several codes, and together they span moisture 25–44 %, EC 5.5–107 µS/cm, and temperature 18.4–21.0 °C. Both verify against the decoder PR (#3602) via rtl_433 -r. Happy to adjust the sample spread if you'd prefer.

PR #509, opened 2026-07-13
Wire codes_test.txt/codes_test.json into run_test.py

merbanan:
Until now these dense-corpus fixtures were purely documentation-by- convention (see AGENTS.md in the main repo): verifying them meant manually piping codes_test.txt through `rtl_433 -y` and diffing by hand. Adds a second pass in run_test.py that does this automatically -- for each codes_test.txt, run every {N}<hex> line through `-y` (respecting the directory's own `protocol`/`ignore` files, same as the .cu8 path) and diff against the matching line in codes_test.json.

### DIFF
--- a/bin/run_test.py
+++ b/bin/run_test.py
@@ -31,13 +31,54 @@ def run_rtl433(input_fn, protocol=None, config=None, rtl_433_cmd="rtl_433"):
     return (out, err, p.returncode)
 
 
+def run_rtl433_y(code, protocol=None, config=None, rtl_433_cmd="rtl_433"):
+    """Run rtl_433 -y <code> and return output."""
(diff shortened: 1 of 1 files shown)


zuckschwerdt:
There is also `test_codes.txt`? Should those be `codes_test.txt` maybe?

Careful with the multi-row -y codes: e.g. tests/florabest/test_codes.txt with
`{30}49093204/{30}49093204/{30}49093204` is actually
`{30}49093204, {0}0, {30}49093204, {0}0, {30}49093204` as each `/` is a new row and a new code `{..}...` is also a new row.

merbanan:
Fixed the bad codes. I'll wire up more regression tests.

PR #511, opened 2026-07-21
Ensure samplerate file existence and non-empty before reading

SamyAlderson:
Added a check to verify the samplerate file exists and is not empty before reading its contents. This prevents a KeyError when trying to access the samplerate. The fix was verified by running the test suite locally and observing 100% pass rate. This change ensures the code is more robust and handles potential file errors.

### DIFF
--- a/bin/refresh_all.py
+++ b/bin/refresh_all.py
@@ -51,11 +51,16 @@ def convert(root, filename, rtl_path):
         print("WARNING: Ignoring '%s'" % input_fn)
         return
 
-    samplerate = 250000
     samplerate_fn = os.path.join(os.path.dirname(output_fn), "samplerate")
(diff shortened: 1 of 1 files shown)


zuckschwerdt:
Thanks for the idea! Did a broken samplerate file pop up somewhere?
Few notes: the Newline at EOF can stay. Wouldn't an empty file be the same error condition as the `int()` not passing? The rate can be float and have a `k` or `M` unit. I think the new `demod` file is much more at risk of broken/bad content.

SamyAlderson:
Thanks for considering my pull request. I agree that an empty file would likely be the same error condition as the `int()` conversion failing, so I can simplify the check. And you're right that the unit could be a float with a 'k' or 'M' suffix. I'll focus on validating the demod file next, as you suggested.

PR #518, opened 2026-08-18
Add 4 tests

rovocz:
Doorbell Solight 1L43, Generic AliExpress door sensor, Honda FRV and Suzuki SX4


zuckschwerdt:
Please add back the meta data that was deleted from the filenames. I.e. frequency and sample rate.

PR #519, opened 2026-08-23
Fix wh52 battery reporting 2

vgabor99:
Pairs with WH52 battery and conductivity reporting fix https://github.com/merbanan/rtl_433/pull/3668.

WH52 battery voltage calculated from byte 20, not byte 15 (based on empirical observations).
Report battery data according to documented data format (single battery_ok in 0.0 .. 0.1 range).
Renamed conductivity to conductivity_uS_cm for consistency and clarity.

Best merged together with (or just after) the decoder PR. Thanks!

### DIFF
--- a/tests/fineoffset/fineoffset_wh52/gfile001.json
+++ b/tests/fineoffset/fineoffset_wh52/gfile001.json
@@ -1,2 +1,2 @@
-{"time" : "@0.007724s", "model" : "Fineoffset-WH52", "id" : "0058ac", "temperature_C" : 20.400, "moisture" : 44, "conductivity" : 106.758, "battery_V" : 1.560, "boost" : 0, "mic" : "CRC"}
-{"time" : "@0.042564s", "model" : "Fineoffset-WH52", "id" : "005995", "temperature_C" : 18.400, "moisture" : 36, "conductivity" : 5.508, "battery_V" : 1.620, "boost" : 0, "mic" : "CRC"}
+{"time" : "@0.007724s", "model" : "Fineoffset-WH52", "id" : "0058ac", "temperature_C" : 20.400, "moisture" : 44, "conductivity" : 106.758, "battery_ok" : 1, "battery_pct" : 86.667, "battery_mV" : 1560, "boost" : 0, "mic" : "CRC"}
+{"time" : "@0.042564s", "model" : "Fineoffset-WH52", "id" : "005995", "temperature_C" : 18.400, "moisture" : 36, "conductivity" : 5.508, "battery_ok" : 1, "battery_pct" : 100.000, "battery_mV" : 1620, "boost" : 0, "mic" : "CRC"}
--- a/tests/fineoffset/fineoffset_wh52/gfile001.json
+++ b/tests/fineoffset/fineoffset_wh52/gfile001.json
@@ -1,2 +1,2 @@
-{"time" : "@0.007724s", "model" : "Fineoffset-WH52", "id" : "0058ac", "temperature_C" : 20.400, "moisture" : 44, "conductivity" : 106.758, "battery_ok" : 1, "battery_pct" : 86.667, "battery_mV" : 1560, "boost" : 0, "mic" : "CRC"}
-{"time" : "@0.042564s", "model" : "Fineoffset-WH52", "id" : "005995", "temperature_C" : 18.400, "moisture" : 36, "conductivity" : 5.508, "battery_ok" : 1, "battery_pct" : 100.000, "battery_mV" : 1620, "boost" : 0, "mic" : "CRC"}
+{"time" : "@0.007724s", "model" : "Fineoffset-WH52", "id" : "0058ac", "temperature_C" : 20.400, "moisture" : 44, "conductivity_uS_cm" : 106.758, "battery_ok" : 0.933, "battery_mV" : 1580, "boost" : 0, "mic" : "CRC"}
+{"time" : "@0.042564s", "model" : "Fineoffset-WH52", "id" : "005995", "temperature_C" : 18.400, "moisture" : 36, "conductivity_uS_cm" : 5.508, "battery_ok" : 1.0, "battery_mV" : 1620, "boost" : 0, "mic" : "CRC"}
--- a/tests/fineoffset/fineoffset_wh52/gfile002.json
+++ b/tests/fineoffset/fineoffset_wh52/gfile002.json
@@ -1,2 +1,2 @@
-{"time" : "@0.011008s", "model" : "Fineoffset-WH52", "id" : "005cb5", "temperature_C" : 20.700, "moisture" : 25, "conductivity" : 56.172, "battery_ok" : 1, "battery_pct" : 93.333, "battery_mV" : 1580, "boost" : 0, "mic" : "CRC"}
-{"time" : "@0.042116s", "model" : "Fineoffset-WH52", "id" : "005b45", "temperature_C" : 21.000, "moisture" : 34, "conductivity" : 25.508, "battery_ok" : 1, "battery_pct" : 93.333, "battery_mV" : 1580, "boost" : 0, "mic" : "CRC"}
+{"time" : "@0.011008s", "model" : "Fineoffset-WH52", "id" : "005cb5", "temperature_C" : 20.700, "moisture" : 25, "conductivity_uS_cm" : 56.172, "battery_ok" : 0.933, "battery_mV" : 1580, "boost" : 0, "mic" : "CRC"}
+{"time" : "@0.042116s", "model" : "Fineoffset-WH52", "id" : "005b45", "temperature_C" : 21.000, "moisture" : 34, "conductivity_uS_cm" : 25.508, "battery_ok" : 0.867, "battery_mV" : 1560, "boost" : 0, "mic" : "CRC"}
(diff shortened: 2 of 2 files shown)


Verstreubulator:
Hi, I wrote the original WH52 decoder (merbanan/rtl_433#3602) and wanted to thank you properly for
merbanan/rtl_433#3668 rather than leave it in the commit log.

Our submission put battery voltage at byte 15. All four of our sensors happened to
sit at nearly the same voltage, so a coincidence looked like a correlation, and we
never checked it against variation within a single unit over time. You found the
real location and wrote the fix yourself. We have since watched the same units over
two months, which is the table further down.

**One thing you may have missed.** The test expectations here were updated on
September 10 in 046e7e2, so the suite is green again and this pull request has
nothing left to do. It was open for about three weeks, and the regression was
failing for most of that time, which is what sent me looking.

**A heads-up, since it touches these same two files.** We have WH52 changes about
to go in: emitting the raw moisture value and the byte 11 range indicator, plus
corrections to comments that our own measurements have since disproved. The tests
side adds the new fields to `gfile001.json` and `gfile002.json`. Nothing there
conflicts with your work now, but I did not want to touch these files without
saying so first. The decoder pull request will say plainly that the byte 20 decode
and the `conductivity_uS_cm` naming are yours.

**Two things you said you wanted, which I think I can supply.**

You wrote that you wondered whether byte 15 is truly constant, and planned to look
into it as time allowed. I have captures of the same four units from July and from
September, two months apart and 60 to 100 mV of discharge:

| unit | byte 15 Jul → Sep | byte 20 Jul → Sep | battery |
|---|---|---|---|
| nw | 0x51 → 0x51 | 0x4f → 0x4a | 1580 → 1480 mV |
| se | 0x54 → 0x54 | 0x51 → 0x4c | 1620 → 1520 mV |
| sw | 0x52 → 0x52 | 0x4f → 0x4b | 1580 → 1500 mV |
| ne | 0x52 → 0x52 | 0x4e → 0x4b | 1560 → 1500 mV |

Byte 15 did not move on any of them. Byte 20 fell on every one, by 60 to 100 mV.
That does not rule out your insertion-time-voltage idea, since none of these had a
battery change in between, but it does say byte 15 is stable across ordinary
discharge. The raw frames are in the repository if you want to check the decode
rather than take my word for it.

You also wrote that the battery depletion curve needs field data. I have Home
Assistant history for eight units going back to early August, about 125,000 state
rows including battery voltage, and it keeps accumulating. It is yours if it is
useful, in whatever shape helps: raw export, per-unit voltage against time, or just
the points where a unit stopped reporting. Say what would be useful and I will
produce it. I would rather hand you data than opinions, given how much of mine I
have had to withdraw.

**The one question only you can answer.** We spent a few days characterising the
sensor and wrote it up here:

https://github.com/Verstreubulator/wh52-characterization

Byte 11 turns out to be an auto-ranging gain indicator, taking ten distinct values
between 5 and 10,000 µS/cm. It tracks reported conductivity monotonically, up as it
rises and down as it falls, in all but one frame of 798, and it does not rescale the
transmitted value.

What makes it worth exposing is that it cannot be recovered from the conductivity
figure. The ranges overlap: the same sensor reported 4,870 µS/cm in one range and
4,700 in the next, so two readings showing the same conductivity can have come from
different gain states, and nothing else in the frame says which. The one exception
to the monotonic behaviour is also interesting — the indicator changes while the
conductivity value sits clamped at 10,000 — but that is a single frame we cannot
check, so I would not build anything on it.

**All of that is from 915 MHz units in one garden: eight for the per-byte work,
four for the per-unit figures.** You mentioned on merbanan/rtl_433#3636 that you
have the 868 MHz variant, which makes you the only independent check I know of.
Does byte 11 behave the same way for you?

The repository has the raw frames and 74 raw IQ captures, so nothing has to be
taken on trust, and a script that recomputes every published figure from the
published data. There is also an errata file listing what we got wrong, which is
longer than I would like. The byte 15 mistake is in it, along with a permittivity
result we published and had to withdraw yesterday.

Corrections and additions are not just welcome, they are needed. I have found it
genuinely hard to build this accurately on my own. Most of what is in that errata
file we caught late, by recomputing our own published numbers rather than by being
careful in the first place, and several of the things we withdrew were checkable
against data that shipped in the very first commit.

I am a homeowner with a software radio, not someone with expertise in this kind of
work. That is the main limitation here, more than the equipment. So any
verification, correction or addition would be genuinely valuable, including telling
me something is wrong without supplying the fix. An independent reading of the same
data would be worth more to this than anything else I can do to it myself.

Which brings me to the actual offer. That repository is currently one person's
measurements of one batch of sensors in one garden, and it says so. If you put 868
MHz readings into it, that stops being true, and the authorship should stop saying
it too.

So: I would rather it were ours than mine. In practice that could mean whatever
suits you — write access and your name on it as an author, moving it to a shared
organisation so neither of us is the landlord, or you keeping your own repository
and the two cross-linking. I do not have a preference beyond wanting the credit to
match who did the work. It is CC BY 4.0 for the documents and GPL-2.0-or-later for
the parser, so nothing about that is encumbered.

There is no obligation in this. Maintaining someone else's project is a real cost,
and "here is some data, do what you like with it" is a perfectly good answer, as is
no answer. But you found an error in our work and fixed it yourself, which is more
than most people would have done, and it seems wrong to keep the byline to myself
after that.

Happy to send raw frames in any format that helps, either way.

Thanks again for the fix.

---

*Edited 2026-09-11. The byte 11 paragraph originally said the indicator "steps monotonically" and that the underlying count "runs continuously across a transition". Checking those properly: it is monotonic in all but one frame of 798, and we have frames inside only one transition, so the continuity claim was not supported by our own data. The reason to expose the field is the range overlap, which does not depend on either. Also corrected: our readings come from eight units across two batches, not four of one. Added: the byte 15 comparison, and an offer on the repository.*

vgabor99:
> One thing you may have missed. The test expectations here were updated on
September 10 in https://github.com/merbanan/rtl_433_tests/commit/046e7e2b0d13fa3df92411cbcebd26137cc48dc9, so the suite is green again and this pull request has
nothing left to do.

I have missed that indeed, thanks for pointing it out. Closing this PR.

> You wrote that you wondered whether byte 15 is truly constant, and planned to look
into it as time allowed. [...] That does not rule out your insertion-time-voltage idea, since none of these had a
battery change in between

After thinking it through, my own measurements rule out the insertion-time voltage theory. I made my analysis swapping 7 batteries across 1.282...1.617 voltage range. Byte 15 remained constant in both sensors.

| V | data | device |
|---:|---|---|
| 1,282 | `a2007713a290195330009516113c938d855a7c393f0998860` | `007713` |
| 1,296 | `a2007713c2901a6030009716113c938d855a7c3940098aa90` | `007713` |
| 1,327 | `a2007713c2901a5c30009616113c938d855a7c394109f10c0` | `007713` |
| 1,379 | `a200771342901a6130019b16113c938d855a7c394409cd760` | `007713` |
| 1,402 | `a200771382901a5b30009516113c938d855a7c394509d1ae0` | `007713` |
| 1,443 | `a200771362901a5d30009616113c938d855a7c3947090cce0` | `007713` |
| 1,617 | `a200771302901c743006ac16113c938d855a7c3950095bfb0` | `007713` |
| 1,282 | `a200846ae281122530008a165444938c86d07bed3e0992480` | `00846a` |
| 1,296 | `a200846aa281122930008e165444938c86d07bed3f096ae90` | `00846a` |
| 1,327 | `a200846a2281133130008b165444938c86d07bed4109878e0` | `00846a` |
| 1,379 | `a200846aa281122530008a165444938c86d07bed440964e00` | `00846a` |
| 1,402 | `a200846a8281122930008a165444938c86d07bed45099bfc0` | `00846a` |
| 1,443 | `a200846ae281122530008a165444938c86d07bed47096a290` | `00846a` |
| 1,617 | `a200846ac2811226300089165444938c86d07bed4f09c56c0` | `00846a` |

> Does byte 11 behave the same way for you?

From the measurements I made for the battery work, I can't tell. I only have 14 data points (2 devices x 7 batteries), measured in ~30 minutes with no variation in soil humidity. All reported range 1 (byte 0x16).

> So: I would rather it were ours than mine

That's a flattering offer but I have to decline. This is related to my interests only tangentially, I don't have this level of commitment. It is also way beyond my knowledge. I have read your analysis in https://github.com/Verstreubulator/wh52-characterization in awe. Sorry!

> you have the 868 MHz variant, which makes you the only independent check I know of

I would be happy to gather reference data (raw or decoded) with my EU sensors. I have 3 and just ordered 2 more, so most likely I will end up devices from different batches too. If you are interested, please let me know what variations of data you need captured. E.g.
 - air / dry soil / damp soil / saturated soil / water
 - salt solution of different levels
 - different temperatures (indoors / outdoors / fridge / freezer)
 - different battery levels (NiMH / Alkaline / Lithium)
 - different soil types (sand / topsoil / compost)

Again, thanks for your work!

zuckschwerdt:
Manually merge by accident. Thanks for the PR!

Verstreubulator:
@vgabor99 Thank you for taking the time to run that. Seven batteries across 1.28 to 1.62 V is a much better test than the one I had, and it settles byte 15: it does not record voltage, at insertion or otherwise. Your frames also show byte 20 rising at every step on both units, which is a useful independent confirmation of your decode. I will add your result to the repository's record, credited to you.

Your table also answers two of our open questions without any further work. Byte 14 is `0x93` on both of your units, as it is on all eight of ours, so it appears to be a constant across both bands rather than per-unit data. Byte 21 is `0x09` on both, which fits the batch pattern we see here (`0x08` on our `0x005xxx` units, `0x09` on our `0x007xxx` units). If one of your new sensors arrives from a different batch, a single frame from it would show whether that pattern holds.

I understand about the commitment, and thank you for the kind words about the repository. Since you offered, here is a short list. Only the first test matters much; please treat the second as optional, and skip anything that is inconvenient.

**1. Salt series, one sensor (about 30 minutes, mostly waiting).** This is the byte 11 question.

- Start with a clean, dry sensor and place its blades in a glass of distilled or demineralised water (the kind sold for steam irons). This is where you could help us most: our series started from tap water at about 320 µS/cm, so we have no frames below that and have never caught the first range change. If distilled water is not to hand, tap water still answers the main question.
- Leave rtl_433 logging for the whole session.
- Every two or three minutes, stir in some salt. At first use only a few grains at a time, because the low end (below about 1,000 µS/cm) is the part we are missing; a few steps between roughly 20 and 300 would be ideal. After that, larger pinches that roughly double the reading are enough to reach the 10,000 µS/cm ceiling. The amounts do not need to be exact, and there is no need to take notes, because each frame carries its own conductivity reading.
- Afterwards, rinse the sensor in clean water and let it dry before it goes back into soil. (Here, an unrinsed probe carried a film that read 45 to 75 µS/cm on its own.)

What would help is the raw frames from the session, in the same form as your battery table.

**2. Air and water, each sensor (a few minutes, optional).** One or two frames with the sensor held in open air, away from other objects (we found that nearby surfaces shift the air reading), and one or two with the blades in plain water. This gives each unit's two end points for the raw moisture value, which would let us compare the 868 MHz units with ours. If the new sensors come from a different batch, the easiest time to do this is before they go into soil.

The other ideas on your list (temperature, battery chemistry, soil types) are all interesting, but I do not think they are worth your time at this stage.

Please attach files wherever is easiest, either here or as an issue on the repository, and whenever it suits you. There is no hurry.

Thanks again.

